PR #54, opened 2015-02-09
LOCALIZATION: add the option of adding default text in localization

karliwalti:
I added the option to provide alternative text in addition to the key to the localization table. The benefit is that if nonverbose keys are used, in case the localization entry cannot be found, still a meaningful string is displayed.


mgorges:
I am wondering if we should instead add support to EMBED files in modules, so the module could ship the required localization file?

karliwalti:
don't have a strong feeling about this, other than the potential risk for
conflicts between an embed in the app and the one in modules.
I just based the example on other lambdanative code that allows to provide
default values.

mgorges:
I was about to press merge when I re-read the code again. If we were to switch to a key based lookup mechanism, not replacing the key string for `(fx= local:index 0)` doesn't make sense. Will discuss with the team tomorrow and then get back to you.

PR #55, opened 2015-03-19
Horizontal Value Picker

jwharrin:
Added: 
Modules/ln_glgui/horizontalvaluepicker.scm

Modified:
Modules/ln_glgui/ln_glgui.scm
-added horizontalvaluepicker.scm


jwharrin:
Created new request without the garbage that was accidentally commited in this one.

PR #62, opened 2015-05-15
scheme tcp repl app and example common lisp client

rick-monster:
Allows to call scheme from common lisp over a tcp port


mgorges:
Thanks for the pull request. Before I merge it two quick questions:
a) What does the repl-client.lisp file do? and b) Why did you place it outside the application folder?

rick-monster:
ah - so that's a repl client that allows to eval some scheme stuff from inside emacs slime (common lisp environment).

PR #76, opened 2015-09-28
Added --without-libssh2 as well as earlier RTMP and IDN modifications…

emacstheviking:
Having removed libraries RTMP and IDN from the curl build, it was also necessary to remove SSH2 as well. This now means it builds out of the box on Yosemite 10.10.5.

### DIFF
--- a/libraries/libcurl/make.sh
+++ b/libraries/libcurl/make.sh
@@ -10,7 +10,7 @@ if [ ! $SYS_PLATFORM = $SYS_HOSTPLATFORM ]; then
   EXTRACONF=--host=$SYS_ARCH
 fi
 
-package_configure $EXTRACONF --disable-ldap --disable-shared --enable-static --without-librtmp --without-libidn
+package_configure $EXTRACONF --disable-ldap --disable-shared --enable-static --without-librtmp --without-libidn --without-libssh2
 
 package_make
 asserterror $? "compilation failed"


emacstheviking:
(curl-GET) .........lovely!

I am rethinking my idea of a mini HTTP server... I want to be creating
lambdanative code NOT tools... I am going to go back to my current
"simple-as-possible" REPL code and finish it...

Goodnight all.
:)
Sean

emacstheviking:
Just when I was starting to cruise...............

 => compiling payload module..

NDK_MODULE_PATH=/Users/seancharles/.lambdanative/tmp.UWVNm9
/Users/seancharles/Documents/RUNTIMES/android-ndk-r10e/ndk-build --silent
APP_ABI=armeabi

file: compiled magic version [522] does not match with shared library magic
version [521]

[armeabi] Compile thumb  : payloadshared <= bootstrap.c

[armeabi] SharedLibrary  : libpayloadshared.so

jni/bootstrap.c:54: error: undefined reference to 'ffi_event'

jni/bootstrap.c:59: error: undefined reference to 'ffi_event'

collect2: error: ld returned 1 exit status

make[1]: *** [obj/local/armeabi/libpayloadshared.so] Error 1

ERROR: failed with exit code 2

BUILD FAILED - configure with verbose option for more information

make: *** [all] Error 1


Tomorrow!

mgorges:
@emacstheviking I have a feeling you are trying to compile the tcprepl for Andoid. This will fail as we currently only support GUI apps for most mobile platforms, as the `eventloop` is required to process events. Try DemoCube for android and see if that builds properly. If it doesn't  would you mind filing a new issue with more details for this? - the pull request is already closed as I merged it. Thanks.

emacstheviking:
Ok.... I was convinced I had this already BUT what I did do was to copy the
tcprepl code into the DemoRedSquare... I have already managed to change the
colour and rounding of the square from both a telnet connection and my
in-progress emacs support code so I know it can works.

I will revisit my DemoRedSquare hacked code and see how I get on, I am sure
this is just "learning curve blues", for educational purposes I am going to
add "-x" to the bash process and see what comes out as the "magic" thing is
a mystery, no pun intended.

I will build a brand new UI application, put in my tweaked repl code and go
from there. Despite all these apparent setbacks, I am pretty damned excited
about this platform and for a nearly 50 years old hacker, that's saying
something, at least for me anyway!

Again, thanks for your help.
Sean.

emacstheviking:
Matthias,

I hope you don't mind me contacting you this way but I have encountered a
problem, unrelated to my previous one, but it has happened several times
now and requires me to reboot my iMac and that's a little odd.

*Every time* I use the ./scripts/lamabdanative utility to create a blank
project like so:

$ ./scripts/lambdanative create repl3 gui

The resulting project fails when I try to configure it:

$ ./configure repl3
./configure: fork: Resource temporarily unavailable

The only way I could avoid this was to manually copy a project not using
the script! I have spent a few hours now. I maxxed out the file handles and
process handles settings and yet it still happens, my machine has plenty of
handles! Even after copying manually the fork problem remains.I 've read
the script over and over but I can see no logical reason why it happens at
all.

$ ulimit  -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
file size               (blocks, -f) unlimited
max locked memory       (kbytes, -l) unlimited
max memory size         (kbytes, -m) unlimited
open files                      (-n) 65536
pipe size            (512 bytes, -p) 1
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 1064
virtual memory          (kbytes, -v) unlimited

I always end up having to reboot my machine to clear the error and it only
happens after running that script...is this a problem you have seen before,
or even can create on the Yosemite machine you have. It is very perplexing
indeed.

All the best.

Sean Charles.

emacstheviking:
$ cat configure
# Automatically generated. Do not edit
here=/Users/seancharles/Documents/RUNTIMES/lambdanative
cd /Users/seancharles/Documents/RUNTIMES/lambdanative
SYS_PATH=$here ./configure $@
cd $here

This is the file causing me trouble.... isn't this recursive and going to
run out of handles pretty quick? The variable "here" is put into SYS_PATH,
then that is used to call configure but that's the script we are all ready
in and so it goes around and around again.... is this right?

Where is the *real* configure... is it that the "./scripts/lambanative
create" has done something interesting OR that I didn't configure it
properly and that has caused it to create this endless looping script!

Here is my ./scripts/lamabdanative configuration from the top part of the
file.....

#!/bin/sh
# LambdaNative initialization script to be placed in system path

# Override this with actual lambdanative location
LAMBDANATIVE=/Users/seancharles/Documents/RUNTIMES/lambdanative
if [ "X$LAMBDANATIVE" = "X" ]; then
  echo "ERROR: this script has not been configured."
  exit
fi

I am going to look into this.... everything was fine before I used this
script LOL!

Sean

emacstheviking:
HI,

I did a "git diff configure" and it was very apparent that the
configuration script had indeed been overwritten...I am guessing that I
have not set the correct variable in the ./scripts/lambdanative script
perhaps?

A quick "git checkout configure" has restored law and order to the process
and I can build once again. I am going to put "-x" in the lambdanative
script and run it again and see what it did, and I know I can fix the fubar
afterwards....maybe another pull request on my fork soon! MAYBE!!

Thanks,
Sean.

mgorges:
Hi Sean,
the lambdanative script is supposed to be used when you have your own code and the SDK in different cases, such as a [system-wide](https://github.com/part-cw/lambdanative/wiki/System-wide-installation) installation of LambdaNative, where you don't have write access to the lambdanative folder. E.g. I have the SDK in `/Users/mg/Documents/Programming/lambdanative` while my own apps/modules etc are located in `/Users/mg/Documents/Programming/parts`

The script should probably check that if the current directory and the SDK directory are identical and throw an error instead of overwriting the configure script. I'll see if I can come up with a patch.

Thanks Matthias

emacstheviking:
Matthias,

I came to the same conclusion after much head scratching! I mean, it
doesn't say not to microwave your poodle to dry it off does it?!

It was time well spent though because I now have a much better internal
picture of what happens when I type "./configure ..." and "make" which
always helps.

https://github.com/part-cw/lambdanative/pull/77

I did the same!

mgorges:
@emacstheviking  sorry, I fixed this about 3 hrs ago and committed that change. Your patch conflicts and solves the same problem in the same way so just refused it.

PR #77, opened 2015-09-29
Added message and exit if script is run from the installation folder

emacstheviking:


mgorges:
Sorry, going to refuse this - already fixed in 0807858

PR #80, opened 2015-11-04
when i tried to compile scheme uSquish program compile give me the at…

javedak09:
scheme issue
…tached error

when i tried to compile scheme uSquish program compile give me the attached error

### DIFF
--- a/scheme issue
+++ b/scheme issue
@@ -1,0 +1,1 @@
+when i tried to compile scheme uSquish program compile give me the attached error


mgorges:
@javedak09 Please file issues here: [New Issue](https://github.com/part-cw/lambdanative/issues/new) instead of making a pull request. The error message tells you exactly what you are missing **ERROR: required tool cmake not found**. 

Please install _cmake_ on your system and then try the built again. If you use macports you can do `sudo port install cmake`, if you use homebrew the command is something like `brew install cmake`

PR #90, opened 2016-01-23
MODULES:LN_GUI: adds scaling to pixmap

karliwalti:
This change enables scaling images with a single factor.
It is useful for switching between different screen sizes.

In planning: defining one scaling measure and automated scaling of other based while maintaining aspect ratio (let me know if this is desirable)


mgorges:
@karliwalti, would you mind issuing two separate pull requests in the future, if you want two features changed? The WAVEOUTPUT change needed some discussion, which delayed the LN_GLGUI change a bit. Thanks Matthias

karliwalti:
Sorry, that was my intention. I did only submit one pull request on
LN_GLGUI. the WAVEOUTPUT was not even pushed yet.

PR #103, opened 2016-03-04
Audio mixing support for various platforms

VincentToups:
Hi Lambdanative folks,

This pull request contains code to add audiomixing support for Android, port audio and IOS devices. The port audio and android versions are tested and working, but the iOS changes, while simple, have not been tested on my device - I hope to get a developer's Macbook sometime next month and then we can resolve this issue.

I _believe_ that windows already supported audio mixing but have no idea about Blackberry. 

I've tried to separate the iOS specific code in separate commits, so it should be easy to revert those in particular if I have made a mess. Let me know if I can help with any other issues - I am busy, but I really like the project and want to keep it active!


mgorges:
@VincentToups: From a coursory glance this looks excellent! I'll review and test the code on a few platforms tomorrow. My only comment right now is, that the piece in db520b8 regarding the "audiofile_count-1" is a bit messy but we can fix it after pulling.

mgorges:
Also for the person merging the pull request 45bf190 removes freebsd in line 344 of `modules/audio/audiofile.scm` which needs to be added back in.

VincentToups:
By the way, it just occurred to me that the MediaPlayerHelper inner class
should be static.

Issue #117, opened 2016-07-04
ln_repl questions


fungiblecog:
I have a couple of questions regarding the ln_repl module, which appears to work differently (with fewer features) from the repl code in the gambit sources.

1) It does not allow some special forms  such as "include" and "define-macro" to be used. Is this by design? If so why? It is a small modification to make it a top-level repl and  enable these.

2) When an error is encountered the repl does not drop into a nested debug-repl. Again I wonder why?

Thanks

mgorges:
@fungiblecog we primarily use modified versions of [DemoConsole](../blob/master/apps/DemoConsole/main.scm) if we need a local REPL for testing, while for remote REPL-like features we use the [rupi](../wiki/Index-of-Module-rupi) module. Also we don't use macros internally. Hence, if you wanted to contribute some modifications to the [ln-repl](../blob/master/modules/ln_repl/ln_repl.scm) module we would welcome a pull-request. Many Thanks!

fungiblecog:
Ok, I will submit something when I'm happy with it.

fungiblecog:
I've been working on enhancements for the repl module on linux and it's working fine so I thought I'd test on some other platforms. I tried to get ln_repl working on the android emulator, but when i telnet i immediately get a 'disconnected by server' error. Before I investigate my setup can you tell me - has the ln_repl module been confirmed as working on Android?

clpetersen:
I just tried the current ln_repl on Android (real device, not emulator).. it works fine, but needed permissions to be accessed remotely. Added in commit f181a8e

fungiblecog:
Great. Now confirmed also working on the android emulator on linux.

PR #120, opened 2016-07-09
Ln_repl define global macros

fungiblecog:
Another enhancement as per issue #117. This one makes all the globally defined macros (in global-macros.scm files) available at the repl when you connect. It's a bit more involved than the previous change as I changed the make.sh script to create a file *all-global-macros*.scm in the ln_repl module directory. This has all the global macro definitions stashed away so they can be eval'd when a connection is made to the repl.

### DIFF
--- a/make.sh
+++ b/make.sh
@@ -358,6 +358,13 @@ compile_payload()
   if [ -f "$appsrcdir/global-macros.scm" ]; then
     cat "$appsrcdir/global-macros.scm" >> "$globalmacrofile"
   fi
+  # enable all global macros to be visible using ln_repl
+  modpath=`locatedir modules/ln_repl silent`
+  (echo ";; Automatically generated. Do not edit."
+   echo "(define global-macros\n\`("
+   cat "$globalmacrofile"
+   echo "))"
+  ) > "$modpath/*all-global-macros*.scm" 
   #--------
   # step 1: compile and assemble the payload objs
   for lng in $languages; do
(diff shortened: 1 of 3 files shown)


mgorges:
@fungiblecog I'll reject and close this one and will ask you to redo the pull-request for two reasons:
a) 3/4 commits are already included in #119, so please rebase against master
b) your make.sh commit alters files in LN's modules/ln_repl folder. To allow system-wide installations, where the LamdbaNative folder is not user-writeable, this won't work for us. Shouldn't they already be included through the compilation anyway?

I'll discuss with @clpetersen on Monday how to accommodate this in a better way. Sorry Matthias

fungiblecog:
a)  I haven't done a 'pull request' before and didn't realise that the second one would also include the first set of changes. Lesson learned there.

b) In order to make the macro definitions available at run-time (they vanish after compilation) I need to stash the contents of the global-macros.scm files in the cache away into a data structure. The only way I could find to do this was to change the make.sh script to create a file in the ln_repl folder. If you can think of a better way to do this I'd love to hear it. In common lisp 'eval-when' allows code run at compile time to set up this kind of thing, but I don't think scheme has this? Even with 'eval-when' I don't think code running in the module would be able to access a file direct from the cache?

mgorges:
@fungiblecog no problem about a, we all start somewhere. As for b, I'll see if I can get them into the cache somehow automatically, so they are evaluated at runtime. I don't know about `eval-when` but it is mentioned in the gambit source `gambit/lib/syntax-case-prelude.scm`. Will look into it tomorrow.

fungiblecog:
Thanks I'll have a look at that reference as well.

PR #121, opened 2016-07-11
Ln repl define global macros 2

fungiblecog:
Should work this time and a much simpler solution.


mgorges:
@fungiblecog thanks for the patch! - I slightly modified your patch to check if the `$globalmacrofile` exists, which is cleaner than redirecting the error to `/dev/null`.

Issue #122, opened 2016-07-11
Build process doesn't invalidate compiled files when global-macros file changes


fungiblecog:
This threw me when testing #121. When a file of global macros is changed I found that the cache for files built with the old version is not invalidated and I had to do a 'make clean' to get the expected result. This is very much an edge-case because of the way I'm sneaking in a run-time definition of macros into the repl but it's useful to know.

mgorges:
@fungiblecog am I understanding you correctly that if you change `/modules/AAA/global-macros.scm` you'd like to have `/modules/AAA/main.scm` recompiled? Or do you also want other modules recompiled?

fungiblecog:
Sorry this wasn't very clear as it was quite late! I think this is only a problem for the enhanced ln_repl module. If any global-macros.scm file is modified the changes are not picked up by the ln_repl module and are not reflected the next time you rebuild. Basically, I want to rebuild ln_repl anytime there is a change to any global-macros.scm file. The easiest thing would probably be just to force it to rebuild every time since it doesn't add much to compilation time and will guarantee users get what they expect. I've not looked too closely at the build process - Is this something you can easily do or shall I look into it?

mgorges:
Solved with in commit 0aac0e1

PR #124, opened 2016-07-13
LN_REPL: (#122) Modify build process to always re-build ln_repl module

fungiblecog:
This forces a rebuild of the ln_repl module to ensure that global macros are always up-to-date.  It's a bit heavy-handed, but it doesn't add much to the compile time, and is a much simpler fix than looking at the timestamp of every global-macros.scm file.


mgorges:
@fungiblecog Thanks - this should be fine!

PR #129, opened 2016-08-14
Usability enhancements and macro rewrites.

VincentToups:
This pull request includes a few minor changes and a few important changes to shadchen and scran. 

The minor changes first: I modified the default frame rate for iOS apps. There are comments indicating that this setting doesn't seem to make a difference for other users, but on my iPhone 6 it makes for a much, much smoother experience.

I've also added a utility to the queue implementation.

The large changes:

gambit-c is really hard to debug if you use macros defined w/ `define-macro`, which operates on a primitive representation of code which is stripped of line number and file information. Any code called by code in a macro-expansion may lose valuable debugging information.

I've rewritten shadchen and scran to use gambit's internal ##define-syntax form so that syntax-objects and information are preserved. This makes them much more usable.


mgorges:
@VincentToups unless you can show that this won't adversely affect battery life I am not happy with tripling the frame rate - in the future please don't merge multiple independent changes in one pull request.

VincentToups:
Matt,

It probably would be bad for battery life - can you you just kill this one
commit or should I make a new pull request without this change?

Will it work to copy this file into an overly and make the change there?
For my use case, a faster framerate is pretty necessary.

-V

clpetersen:
@VincentToups, looking at your shadchen changes, there are a number of match calls that appear like unit tests; lines 280-294. Is this intentional? Also, the `...` notation seems odd?  Finally, the `display` line at the end of the commit might break on platforms without stdout.  Please refactor the pull request as per @mgorges suggestion and resubmit. Thanks

PR #136, opened 2016-10-19
Scran unit tests + utils.

VincentToups:
This commit adds three unit tests to scran, plus a few more utilities.

### DIFF
--- a/modules/scran/scran.scm
+++ b/modules/scran/scran.scm
@@ -94,7 +94,17 @@
 		   (begin 
 			 (add-to-system-raw! e anti-system))
 		   #f))
-	 anti-systems)))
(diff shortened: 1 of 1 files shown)


VincentToups:
I should have said "commit chain".

mgorges:
Wouldn't `!reset-scran!` better be called `reset-scran!`?

VincentToups:
The extra bang is to emphasize the extremely unusual nature of the
function, which really wouldn't be used in a typical lambdanative project.
I really only added it to support unit tests, which need to break down the
whole system to not have side effects.

More typical usage is to define components and systems once in the entire
run of a game or app and only reset the systems sometimes.

-V

PR #141, opened 2016-11-21
ORIENTATION: add check for inverted orientation (seascape/upsidedown)

fungiblecog:
Current implementation doesn't provide a function to distinguish inverted orientations. Could add a separate check for each orientation but this allows the existing behaviour to be unchanged (ie landscape/portrait is considered to be either of two orientations).


mgorges:
@fungiblecog thanks - can you please also add documentation of the function to the wiki. Thanks!

fungiblecog:
Will do

PR #164, opened 2017-11-24
updrade redcap to latest functions

karliwalti:
mainly, adds support for dealing with repeated forms (available since ~Redcap 6, working reliably since redcap 6.4)


mgorges:
@karliwalti I fixed your broken `make-list-natural` function, which was included in this pull request, as I wasn't able to selectively pull and squash. Doing do I switched the order of arguments, which make more sense that way. Next time can you create two pull requests for separate pieces of code, please?

karliwalti:
@mgorges  Sorry, I really made the redcap request first, then added the list.  I github combining these?

karliwalti:
actually, I did not yet create a push request for the list.scm, but I might accidentally, pushed it to remote after the pull request.

Thank you for fixing the list function!

mgorges:
It was puzzling to me as well, as I thought pull requests were based on commit hashes not remote branches? Anyway, when I was trying to merge I got a notification that the merge request had changed and I would have to re-review it. I didn't find a way to only pull/merge only the first n-1 commits, but if you can point me in the right direction for next time I'd appreciate it. Thanks!

PR #165, opened 2017-11-25
uiform layout procedures

karliwalti:
The changes permit to align text in lists and change the button color individually. The configurations are made so that they are perfectly back compatible


mgorges:
Thanks for the pull request. I squashed and merged the commits. Can you rebase, so that future pull requests only contain the new commits, and not also the 18 before it, which go back to 2016 and are mainly remote-tracking updates? Thanks!

PR #167, opened 2017-11-30
ALIST adds function to insert/replace list item without timestamp

karliwalti:
timestamp is not always needed


mgorges:
@karliwalti can you please also document any new functions/features in the Wiki? Thanks!

PR #168, opened 2017-11-30
Uiformimprovement

karliwalti:
I was hoping creating a pull request from a branch would prevent from showing up unrelated commits from master. did not work.
I was also not able to push individual commits. they are all there now.

Remarks: I could not find any issues with the changes, Older code should not break as I put the default parameters to what it would be without. 

I noticed that `(abs (- (abs y) (uiget 'offset 0) h ))` (getting the position of the element on the gui, `(uiset 'offset value)  `  shoud jump to the element) does not give the correct value in rare cases. As the value is lower than expected it has no consequence for now.

I will add the entries to the wiki


mgorges:
Unfortunately, this is not as simple as I would like it to be and highlights we need documentation for it:
1) If you wanted to only make a pull request for the ALIST commit, you could select `48b69a5` in the head-fork compare dropdown (the right of the four). - The same would work if you wanted to include a number of past commits but not all the way go to the current head. 
2) If you want to separate the next seven commits and send them individually, instead of as a big list, one way is to create a new branch for each one, and **cherry-pick** the commits you want to add to it from your master branch and then make pull requests for each one. However, this is a command line exercise and also a commonly requested feature [isaacs/github/issues/629](https://github.com/isaacs/github/issues/629).

mgorges:
Sorry, what I also don't understand is why the ALIST commit is in here, as I already merged that this morning?

mgorges:
Commit b8552b2 fixes a bug introduced in 2a733a7 which crashes camera in WidgetDemo.sx as `(symbol->string #f)` causes a segfault if 'id is not set. Could you test with that to make sure there are no additional issues, please?

PR #238, opened 2019-05-15
Updating gambit

0-8-15:
Looks like recent versions of gambit have fixed some of the issues wrt. macro expansion.

So far lambdanative still works for me(TM).

### DIFF
--- a/libraries/libgambit/make.sh
+++ b/libraries/libgambit/make.sh
@@ -1,3 +1,9 @@
+# DON'T TRY 4.9.3 gambit issue #384 define-syntax is broken
+# see https://github.com/gambit/gambit/issues/384
+# PKGURL=https://www.iro.umontreal.ca/~gambit/download/gambit/v4.9/source/gambit-v4_9_3.tgz
+# PKGHASH=a3f836f732f367601d0df8bedb8e1324a14c413b
+PKGURL=https://www.iro.umontreal.ca/~gambit/download/gambit/v4.9/source/gambit-v4_9_2.tgz
+PKGHASH=70c2e66cdbaf544afdcfa6d396c37f72ec67e934
 PKGURL=https://github.com/gambit/gambit/archive/v4.8.5.tar.gz
 PKGHASH=08d129513d04970f30302f28cac17c4e21a25a6b
 


mgorges:
I am extremely hesitant to consider merging this as most changes to the Gambit compiler since 4.7.0 have been related to the inclusion of other languages, like Javascript etc, instead of changes to functionality. It might be that our use case (medical software development) is more conservative but the testing about needed to ensure we don't accidentally break something - can't wait to look at Android (heartbeat) - is very high. We have been using 4.8.5 for some projects, as you note by patching the [libgambit](https://github.com/part-cw/lambdanative/tree/master/libraries/libgambit), but there was nothing in terms of gaining performance/features that made me jump on changing to it for everything. 

I am willing to change the `libgambit` piece, so users can chose to go there for their own systems, but making it the new default is not something I am willing to risk at this time, as I don't have the capacity to properly test for it and then get blamed if things break.

mgorges:
The "don't use 4.9.3" comment is useful too - thanks.

0-8-15:
Hi Matthias,

I understand your hesitations.  Especially wrt. the medical
applications.  At the other hand: we plan to target medical apps too.
That's one thing which made lambdanative interesting to us.  (I
enthusiatically informed my collegue here yesterday that we could
actually use lambdanative soon, since the two years old road block would
be gone.)

Frankly, I'm not really interested in those bloat features of recent
gambit.  I'd rather see my two year old issue solved wrt. IPv6.

However a Scheme implementation without hygienic macros is just another
programming language.  With them it's any language you want.

There is a large codebase (roughly 250kLOC in ~100 modules) I'd
like to port[1].  No way without syntax-rules.

Are there any know issue with 4.8.5?  Anything I could do help
testing?  Or maybe merge it into some extra branch?

Best

/Jörg

[1] At the end of the day I'd like to port ball.askemos.org to
lambdanative.  This is essentially the first "smart contract"
environment.  Contrary to those public blockchain things this one was
designed with privacy in mind.  (Data is replicated/audite only with
_user selected_ wittness/notary nodes, no public blockchain to leak
medical data on.)

Am Wed, 15 May 2019 13:17:38 -0700

> I am extremely hesitant to consider merging this as most changes to
> the Gambit compiler since 4.7.0 have been related to the inclusion of
> other languages, like Javascript etc, instead of changes to
> functionality. It might be that our use case (medical software
> development) is more conservative but the testing about needed to
> ensure we don't accidentally break something - can't wait to look at
> Android (heartbeat) - is very high. We have been using 4.8.5 for some
> projects, as you note by patching the
> [libgambit](https://github.com/part-cw/lambdanative/tree/master/libraries/libgambit),
> but there was nothing in terms of gaining performance/features that
> made me jump on changing to it for everything. 
> 
> I am willing to change the `libgambit` piece, so users can chose to
> go there for their own systems, but making it the new default is not
> something I am willing to risk at this time, as I don't have the
> capacity to properly test for it and then get blamed if things break.
>

mgorges:
I personally haven't had any issues with 4.8.5 but we are not using it for most of our production apps either. The platform I am most concerned running into issues with is Android (and maybe iOS too), which is difficult to debug. Would 4.8.5 be sufficient for your needs, or is there something specific in 4.9.2 you need?

mgorges:
I was about to merge a slightly modified version of this, but it breaks the current Android build. [gambit:7d2de2a](https://github.com/gambit/gambit/commit/7d2de2acaea0631d665cb3909fbc0238435cdc89#diff-3253c143da0df6ef33aca44ffdb95b10) removes `___disable_heartbeat_interrupts`, and 
[gambit:62cce77](https://github.com/gambit/gambit/commit/62cce77dcb59cd807d8b0376a405acfc3deab1e6) renames `##set-heartbeat-interval!` to `##set-heartbeat-interval!`, both of which we use for Android.

0-8-15:
This commit
https://github.com/0-8-15/lambdanative/commit/765809f887733cd531794bd3db5061c96a5122c0
should work for both version of gambit.

0-8-15:
Replaced by better version.

PR #240, opened 2019-05-22
Support macro loading per file.

0-8-15:
This patch provides a partial way to specify compile time macros.

Iff there is a file with the same basename and an extension .sch
the latter is loaded into gambit via -e option.

Caveat:

lambdanative does not check the .sch file for modifications like it
does with the .scm files.  Updates to the file do not trigger a
recompile.

A real solution would be to support r7rs style library declarations.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -89,6 +89,7 @@ compile_payload_scm()
       scm_ctgt="$SYS_PREFIX/build/${scm_chsh}.c"
       scm_otgt="$SYS_PREFIX/build/${scm_chsh}.o"
       scm_dirty=no
+      scm_hdr="$scm_path/"`basename $scm_src .scm`.sch
       if [ ! -f $scm_otgt ] || [ ! -f $scm_ctgt ]; then
         scm_dirty=yes
       else
@@ -102,6 +103,10 @@ compile_payload_scm()
           if [ `newersourceindir "$scm_src" "$scm_otgt"` = "yes" ]; then
             scm_dirty=yes
           fi
+	  # This is actually much too less, should test all included/loaded files too.
+          if [ -f $scm_hdr -a `isnewer "$scm_hdr" "$scm_otgt"` = "yes" ]; then
+            scm_dirty=yes
+          fi
 	  # ln_repl module special-case: always re-compile to make global macro changes available
 	  if [ "X$scm_topdir" = "Xln_repl" ]; then
 	     scm_dirty=yes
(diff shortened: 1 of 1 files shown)


mgorges:
I am confused by this comment: `lambdanative does not check the .sch file for modifications like it does with the .scm files.  Updates to the file do not trigger a recompile.`, as it seems you do check for that time and trigger a recompile?

0-8-15:
I found that languages/scm.sh uses `newersourceindir` to check all
*.scm files in the directory of the $scm_src given.  Now that I'm
adding the *.sch file this is not being checked the same way.

Maybe that should be done.

I still wonder, maybe tehre is a better way to accomplish the trick.
Just beyond me.

Am Thu, 23 May 2019 13:31:00 -0700

> I am confused by this comment: `lambdanative does not check the .sch
> file for modifications like it does with the .scm files.  Updates to
> the file do not trigger a recompile.`, as it seems you do check for
> that time and trigger a recompile?
>

mgorges:
The [newersourceindir](https://github.com/part-cw/lambdanative/blob/master/make.sh#L180) test only works for *.scm, as that is the only language that is currently defined in this repository. However, the testing loop could be used instead of line 107 of the patch to get higher coverage of all *.sch files?

0-8-15:
I'm slightly confused which line 107 you referred to.

At the other hand I'm not convinced that the approach to recompile a `.scm` file whenever another `.scm` file in the same directory has changed is a good one in the first place.

The real solution(TM) is IMHO to use r7rs library definitions to resolve dependencies.  That could handle syntax definitions imported from other modules/locations, normal code, larger modules with code in subdirectories etc.  Currently I to have a a look into how this could be done. [black hole](https://github.com/per-gron/blackhole) and [SchemeSpheres](http://www.schemespheres.org/) came up as options.

For the time being, I'd rather not add more complexity to the build process.  The proposed change would support precisely one `.sch` file of the same `basename` as the file to be compiled.

mgorges:
I was referring to that line in your [pull request diff](https://github.com/part-cw/lambdanative/pull/240/commits/cbb9d1eecfebd52886b27706c7ce2368abf25844#diff-22273f1d94f654e2c8e76c51bc402760R107), which is 
> if [ -f $scm_hdr -a `isnewer "$scm_hdr" "$scm_otgt"` = "yes" ]; then`

However, given that you don't wish to expand it to allow wildcard support, I'll review the pull request as is and merge if it doesn't cause obvious issues.

PR #241, opened 2019-05-23
Update to recent gambit.

0-8-15:
Change USE_EXPERIMENTAL in languages/scm.c to switch between gambc
4.7.9 and gambit 4.9.2.

### DIFF
--- a/languages/scm.c
+++ b/languages/scm.c
@@ -8,9 +8,23 @@
 #define DMSG(fmt...)
 #endif
 
+/*
+ * Maybe we better fist #include <gambit.h> and then check that we've
+ * got a known to be good version instead of unconditionally
+ * overwriting it?
+ */
+#define USE_EXPERIMENTAL 0
+#if (USE_EXPERIMENTAL = 0)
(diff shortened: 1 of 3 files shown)


mgorges:
Merged and fixed a few small things in 490038a, including that `#if` doesn't end with `#fi`. I also removed the hardcoded version number, but I don't know it this breaks things if one switches between libgambit and libgambc without scrubbing the cache?

0-8-15:
Am Thu, 23 May 2019 12:34:33 -0700

> Merged and fixed a few small things in 490038a, including that `#if`
> doesn't end with `#fi`. I also removed the hardcoded version number,

Ups.  Did I mess it up with the sugar I added post-test.  I always
tell the people not to.   But than, given the small machine I have to
use for this... I confess, I did. 

> but I don't know it this breaks things if one switches between
> libgambit and libgambc without scrubbing the cache? 

Nah, we better document that this is an either or.  Not scrubbing the
cache has extended my testing time considerably.

PR #249, opened 2019-07-13
Add support for libmagic.

0-8-15:
Not sure that this is the best implementation to use.  I just needed
any.

TODO: Figure out how to rename the file upon package_download.  The
basename "5.18.tar.gz" appears meaningless and may conflict.

### DIFF
--- a/libraries/libmagic/make.sh
+++ b/libraries/libmagic/make.sh
@@ -1,0 +1,76 @@
+PKGURL=https://github.com/threatstack/libmagic/archive/5.18.tar.gz
+PKGHASH=ce734a1cc24bddbfd8b21f4ffdfe721ab74eeed9
+
+package_download $PKGURL $PKGHASH
+package_patch
+
+EXTRACONF=
+
+if [ "$SYS_PLATFORM" != "$SYS_HOSTPLATFORM" ]; then
+    EXTRACONF="$EXTRACONF --host=$SYS_ARCH"
+fi
+
+# pretend libgnurx exists by masking libpcre
+
+test -L $SYS_PREFIX/include/regex.h || ln -s pcreposix.h $SYS_PREFIX/include/regex.h
+
+# Mabye instead of fooling the package, it should be patched
+rmifexists $SYS_PREFIX/lib/libgnurx.a
+rmifexists $SYS_PREFIX/lib/libgnurx
+mkdir $SYS_PREFIX/lib/libgnurx
+( cd $SYS_PREFIX/lib/libgnurx
+  mkdir a
+  cd a
+  $SYS_AR -x ../../libpcre.a
+  cd ..
+  mkdir b
+  cd b
+  $SYS_AR -x ../../libpcreposix.a
+  cd ..
+  $SYS_AR rc ../libgnurx.a a/*.o b/*.o
+)
+
+rm -r $SYS_PREFIX/lib/libgnurx
+
+if [ `file --version|head -1` != file-5.18 ]; then
+    if [ ! -f  $SYS_HOSTPREFIX/bin/file ]; then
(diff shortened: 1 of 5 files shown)


0-8-15:
It would be a good idea if the `magic` module would make sure the file installed under `$SYS_PREFIX/etc/magic.mgc` was embedded in the executable and found as the default for `magic-load`. I just did not yet understand how that would be done.

libraries/libmagic/make.sh
```diff
@@ -0,0 +1,76 @@
...
+
+rm -r $SYS_PREFIX/lib/libgnurx
+
+if [ `file --version|head -1` != file-5.18 ]; then
+    if [ ! -f  $SYS_HOSTPREFIX/bin/file ]; then
+	./configure --enable-static --disable-shared 'CFLAGS=-fPIC -O3 -DPCRE_STATIC'
+	package_make
+	cp src/file $SYS_HOSTPREFIX/bin || exit 1
+	package_cleanup
+	package_download $PKGURL $PKGHASH
```
mgorges: shouldn't it already be downloaded by this time (as in line 4)?

mgorges:
Embedding of files can be done with the [EMBED](https://github.com/part-cw/lambdanative/wiki/Embed) functionality, but I don't think this works for any path outside the app or module directory, as that is not what [apps/packtool/main.scm](https://github.com/part-cw/lambdanative/blob/master/apps/packtool/main.scm) does.

libraries/libmagic/make.sh
```diff
@@ -0,0 +1,76 @@
...
+
+if [ "$SYS_PLATFORM" != "$SYS_HOSTPLATFORM" ]; then
+    EXTRACONF="$EXTRACONF --host=$SYS_ARCH"
+fi
+
+# pretend libgnurx exists by masking libpcre
+
+test -L $SYS_PREFIX/include/regex.h || ln -s pcreposix.h $SYS_PREFIX/include/regex.h
+
+# Mabye instead of fooling the package, it should be patched
```
mgorges: Not quite sure about this either. Why not include that library if you depend on it?

libraries/libmagic/make.sh
```diff
@@ -0,0 +1,76 @@
...
+
+package_make
+
+# install
+
+cp src/.libs/libmagic.a $SYS_PREFIX/lib
+cp src/magic.h $SYS_PREFIX/include
+test -f src/file.exe && cp src/file.exe $SYS_PREFIX/bin
+test -d $SYS_PREFIX/etc || mkdir $SYS_PREFIX/etc
+cp magic/magic.mgc $SYS_PREFIX/etc
```
mgorges: If you need this file embedded couldn't you compile it once and then include it in the module with an EMBED file? - Also does it need to be in the `/etc` folder?

libraries/libmagic/make.sh
0-8-15: Yes it should reading `package.sh` again I now see that this `package_download` might be better replaced by a `package_unpack`.  I just was not aware of the latter being available as a separate call.

libraries/libmagic/make.sh
0-8-15: Two reasons:

1. At least I need `pcre` anyway and I see limited reasons to have two libraries implementing the same functionality.

2. I'd guess there might be license compatibility issues when linking against a GPLed library and `libgnurx` is AFAIK.

libraries/libmagic/make.sh
0-8-15: Not sure.  I don't really know where to put it.  As you said: embedding might not be available from within the library directory.  Hence we'll have to install it somewhere to be available for embedding later.

PR #252, opened 2019-08-15
Compare string in sh/test with '=' not '=='

0-8-15:


mgorges:
https://linuxize.com/post/how-to-compare-strings-in-bash/ suggests to use "==" for [[ command for pattern matching

PR #257, opened 2019-08-29
Handle compilation and permissions for ANDROIDAPI>22.

0-8-15:
Since Marshmallow permissions are not automatically granted as
specified in the Manifest.

As a consequence LN apps die when accessing the SDCard, e.g. when
unpacking embedded files.

This patch adds a fragile hack to enable an ifdef-alike trick to
exclude code portions when targeting older APIs and uses it to request
the permission at startup.  Plus create the system-directory if it
does not already exists.

### DIFF
--- a/loaders/android/bootstrap.java.in
+++ b/loaders/android/bootstrap.java.in
@@ -111,7 +111,7 @@ public class @SYS_APPNAME@ extends Activity implements @ANDROID_JAVA_IMPLEMENTS@
     setContentView(mGLView);
     mSensorManager = (SensorManager)getSystemService(Context.SENSOR_SERVICE);
 
-    /* Included @IF_ANDROIDAPI_GT_22@
+    @IF_ANDROIDAPI_GT_22@
     if (getApplicationContext().checkCallingOrSelfPermission(android.Manifest.permission.WRITE_EXTERNAL_STORAGE)
         !=  android.content.pm.PackageManager.PERMISSION_GRANTED) {
         requestPermissions(new String[] { android.Manifest.permission.WRITE_EXTERNAL_STORAGE }, 1);
--- a/loaders/android/bootstrap.java.in
+++ b/loaders/android/bootstrap.java.in
@@ -75,6 +75,7 @@ public class @SYS_APPNAME@ extends Activity implements @ANDROID_JAVA_IMPLEMENTS@
   //Variable declarations needed for modules, e.g. gps
   @ANDROID_JAVA_VARIABLES@
 
+  int idle_tmScheduleRate = 250;
   Timer idle_tm = new Timer();
   TimerTask idle_task = new TimerTask() {
     public void run() {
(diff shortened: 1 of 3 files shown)


mgorges:
The conditional substitution of comments start and end tags is an interesting workaround! It is the `requestPermissions` which needs the 'Android Support Library' of at least version 24.1.0, so we'll need to make it mandatory, it previously was not. Also I thought the minimum supported API level for support libraries  is API level 14, in which case we wouldn't need your `ifdef` like construct?

0-8-15:
Am Thu, 29 Aug 2019 10:49:43 -0700

> The conditional substitution of comments start and end tags is an
> interesting workaround! It is the `requestPermissions` which needs
> the 'Android Support Library' of at least version 24.1.0, so we'll
> need to make it mandatory, it previously was not.

That's strange to me.  I tried to use parts of the 'Android Support
Library' but found them not supported on my system at all.  Thus I'm
confused that I ended up using anything from it.  I never too (on
purpose) any steps to install any code beyond what was installed
follwing the LN documentation.  Who knows how it became installed?  Is
it actually?  How would I tell?

Granted I'm struggling a bit to gather what is required/use on
Android.  I have only so much experience with it.

Having said that, I'd rather try to get along without that library if it
makes sense.  However maybe it does.  After all supporting older API
versions looks so far easier to me with it than without.

How would I actually replace `requestPermissions` to get an equivalent
dialog without it?  Do you know?

> Also I thought the> minimum supported API level for support libraries
> is API level 14, in which case we wouldn't need your `ifdef` like
> construct?

I'd appreciate if we would not need /that/ `ifdef` hack at all!  It's
fragile.  It could at least be done better (having propper replacements
for BEGIN_JAVA_IF_* end END_JAVA_IF_* plus {BEGIN,END}_IFNOT_* and
*_XML_IF*_* for conditional replacements in the Manifest.xml etc.) to
still have syntax highlighting in the source.  But that would be a
lot of additional code, which in turn would need to be maintained
(I'd bet this ends in chaos, if actually done).

As a second point I'm surprized that the mininum API is level 14.
Right now I don't recall where I learned that it would be level 19, but
recall (a recall which could be wrong) that I choose that initially due
to having read it /somewhere/ in the LN documentation.

Nevertheless the `ifdef` like hack is IMHO so far required to enable
/exclusion/ of code only valid in API level 23 and /above/.  Hence I
don't see the point how this could relate to level 14.  Those methods
would not be available there either.

0-8-15:
The [other pull request](https://github.com/part-cw/lambdanative/pull/256/commits) now includes the latest change, which preserves syntax highlighting at least.

0-8-15:
As it I might understand what you meant by "minumim API level 14", i.e., I could simply abstain from using "API level 23 and above": recent Android versions appear to kill my app while it is supposed to supply a ongoing service.  The solution - which still requires further testing to build confidence, seems to be to go through more recent APIs (i.e., at least level 23) to convince Android to leave it running.  That's the purpose the [ForegroundService](https://github.com/part-cw/lambdanative/pull/256) attempts to achieve.

This pull request is only a spin-off of the changes required, which need to go into LN core parts.

As a side effect I expect goggle to to deprecate more aspects of older API, like permission handling, which would eventually force users to change the minimum API level for one reason or another.

mgorges:
You need to target at minimum API 26 to submit to the play store, but your minimum API level can be as low as you like. This is relevant particularly for the global health app space, in which we do some things, so I don't want to raise it beyond what we actually need to. The minimum according to the example would be API 19, so going from 19 to 23 would drop almost 15% of devices currently in use. 

From the documentation it seems that `requestPermissions` is part of the Android Support
Library, but even when I remove it I can still compile so I will merge this request shortly. I'd also prefer to merge this one and then the other one, instead of doing a single large merge, to keep things reasonably easy to figure out when they break.

PR #259, opened 2019-09-05
Name mutex's to ease debugging.

0-8-15:


mgorges:
Could you add this for all places where a mutex is used, not just these two?

0-8-15:
Am Thu, 05 Sep 2019 12:11:50 -0700

> Could you add this for all places where a mutex is used, not just
> these two?


How about "cum tempore".  I'm sitting on several modifications not
ready to commit, trying to fix unrelated issues.  Either I submit
untested untested code (again) or piece by piece.

PR #261, opened 2019-09-06
Tag mutex in ide-repl-pump

0-8-15:
Caveat: This code is untested AND I don't understand the code!!!

Somehow the procedure `ide-repl-pump` is called in recursiv manner,
which will create a fresh mutex.  I don't see which ressource exactly
this mutex is supposed to protect.  Hence I do not understand whether
or not it is correct to actually allocate two different mutexs here.

PLEASE review this with care to ensure we don't create two mutexes to
protec the same ressource and hence fail to protect it at all.


mgorges:
Sorry but I didn’t write or use this module so I don’t think I can comment on expected behaviour. @clpetersen or @fungiblecog might?

PR #263, opened 2019-09-06
Enable use of cond-expand depending on SYS_PLATFORM.

0-8-15:
This change allows to use things like
   (cond-expand
    (android ...code applicable on android targets only...)
    (linux ...code applicable on linux targets only...)
    (else ...code applicable on all other targets...))

NB:

1. According to SRFI-0 `cond-expand` is only mandatory on top level.
2. Most Scheme implementations (i.e., all I know of), including gambit
   allow the use of `cond-expand` everywhere.
3. With this change all uses of `app:android?` at runtime could
   eventually be replaced by compile time code expansion.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -56,6 +56,7 @@ compile_payload_scm()
     else
       scm_opts="(declare (block)(not safe)(standard-bindings)(extended-bindings))"
     fi
+    scm_opts="${scm_opts}(define-cond-expand-feature $SYS_PLATFORM)"
     # support global macro definitions
     if [ -f "${SYS_HOSTPREFIX}/lib/global-macros.scm" ]; then
       scm_opts="${scm_opts}(include \\\"~~lib/global-macros.scm\\\")"


mgorges:
Again not a feature I ever used and from a cursory review it only seems to be mentioned in the ‘ssax’ module and something you recently contributed. What is the performance hit we take when enabling this and does it break anything in terms of recompilation or other default behaviour?

0-8-15:
Am Sat, 07 Sep 2019 16:17:33 -0700

> Again not a feature I ever used and from a cursory review it only
> seems to be mentioned in the ‘ssax’ module and something you recently
> contributed.

cond-expand a.k.a. [SRFI-0](https://srfi.schemers.org/srfi-0/) is the
standard Scheme way do to conditional code expansion.  It's the
'#ifdef' of Scheme.

> What is the performance hit we take when enabling this and does it
> break anything in terms of recompilation or other default behaviour?

Hardly there is a performance hit from simply declaring the feature.

If it was used (e.g., to replace `(if app:android? ...)` the
performance hit is negative by a small amount since cond-expand is
compile time.  The test-and-branch would not exist at all.

PR #250, opened 2019-07-13
Update openssl to 1.1.c

0-8-15:

PR #266, opened 2019-09-24
android: cross compile using clang

0-8-15:


mgorges:
Why do you need `LN_C_compiler0` and `LN_C_compiler` - can't the `ls` find clang itself? Also was `clang` required to build with Android before this addition? If so, do we test for it in `make.sh`?

0-8-15:
Am Tue, 24 Sep 2019 12:18:30 -0700

> Why do you need `LN_C_compiler0` and `LN_C_compiler` - can't the `ls`
> find clang itself? Also was `clang` required to build with Android
> before this addition? If so, do we test for it in `make.sh`?

That's a bit of a hack.  It might be better a situation now, once API
version is in the cache path.  Using `ls ...*-clang` IIRR turned out too
many hits.  But I found only one `gcc`.  To compile with clang, the code
looks for the gcc now and use the correct clang from that dir.  Strange
but does the trick.  Will likely have to be changed with the next ndk,
unfortunately.

mgorges:
So I guess `targets/android/check-tools` needs to also check for clang and the change in #250 will need to be undone?

0-8-15:
Am Wed, 25 Sep 2019 19:16:21 -0700

> So I guess `targets/android/check-tools` needs to also check for
> clang and the change in #250 will need to be undone?

Likely, I'd say.  Dunno.

mgorges:
I briefly tested this on _macOS_ which you don't yet patch and I see a performance difference in compilation [Clang: 3m53.320s, GCC: 4m27.935s]. Don't have an Android device by my desk to test if the code works, but it does compile.

ddunsmuir:
Please note that clang is not present (at least not in the expected location) in older versions of the Android NDK. Using version r11c, the file /Users/ddunsmuir/Library/Caches/lambdanative/android-api19/android-ndk-r11c-arm-19-toolchain/bin/arm-linux-androideabi-clang does not exist and so the build failed. Getting the latest version, r20b solved this issue. I'm not sure what the oldest you can use is, but potentially it is r19:
"As of NDK r19, the toolchains installed by default with the NDK may be used in-place." https://developer.android.com/ndk/guides/other_build_systems

PR #268, opened 2019-09-24
Android: APIVERSION in file names of cache directory and package.

0-8-15:
This should shell you from the confusion developing for several API
versions.  (Or the tidious need to wait for `make scrub` every time.)


mgorges:
I am wondering if the piece in the `SETUP.template` should go to `make.sh` as most users will never retrieve this file once they created their own `SETUP` file. Thus `SYS_PLATFORM_VARIANT` would be undefined which might cause problems with debugging?

0-8-15:
Am Tue, 24 Sep 2019 14:44:08 -0700

> I am wondering if the piece in the `SETUP.template` should go to
> `make.sh` as most users will never retrieve this file once they
> created their own `SETUP` file. Thus `SYS_PLATFORM_VARIANT` would be
> undefined which might cause problems with debugging?

I was wondering where the piece shold go for good.

Would you just move it to the appropriate place?

Best

PR #269, opened 2019-09-25
Android embed apiversion

0-8-15:


mgorges:
How is this one different from #268?

0-8-15:
Am Wed, 25 Sep 2019 16:38:55 -0700

> How is this one different from #268?

Sorry, looks like githug automatically notifying.

Looks like I pull upstream/master into the branch and commited.

The thing, which irritated me, was that the embedding of
`SYS_PLATFORM_VARIANT` into the final `.apk` did not make it into
upstream, did it?  Without I'm likely to confuse both versions, which
is why I'd like it to be the case.

But maybe I just shot myself into the foot with `git`.  The latter is
known to be a risky wappon.  :-/

0-8-15:
This looks nice from a backward-compatiblity focused point of view.

But simply deleting the setting of `SYS_PLATFORM_VARIANT` from the `SETUP` file does not cut it.

This is saying: "works but just for me".  Users not setting  `SYS_PLATFORM_VARIANT` from their `SETUP` will still run into the issues the change should fix.

mgorges:
I don't see the change to `make.sh` but maybe its Github that is screwing things up for me. Should I merge it and then move this to `make.sh` line 954?

> if [ "$SYS_PLATFORM" = "android" ]; then
>   SYS_PLATFORM_VARIANT="-api${ANDROIDAPI}"
> fi

0-8-15:
That's what I'd suggest.

However maybe we should so only if the `SETUP` file as read by
`setstate SETUP` did not already set `SYS_PLATFORM_VARIANT`.

Right now I don't see how this could have come to be anywhere but on my
source tree (and _that one_ should not count here).  But maybe users
knowin what they do want to override the new behaviour.

Am Fri, 27 Sep 2019 22:21:32 -0700

> I don't see the change to `make.sh` but maybe its Github that is
> screwing things up for me. Should I merge it and then move this to
> `make.sh` line 954?
> 
> > if [ "$SYS_PLATFORM" = "android" ]; then
> >   SYS_PLATFORM_VARIANT="-api${ANDROIDAPI}"
> > fi  
>

PR #273, opened 2019-11-11
DEMO: LNjScheme (Demo, Android only)

0-8-15:
This is a demo how a simple Scheme interpreter in Java (derived from Peter Norvig's jScheme) can be used to run arbitrary Jave/Scheme code in either the main UI thread of Android or any thread. Due to Android's requirements wrt. changing views the former is likely the more interesting option.

The main idea behind is to generalize the special cases currently used for `hybridapp`s (and remove the special handling from the build process) while making it easier to use native widgets from lambdanative.  (E.g., I don't like to force the user to go through the hassle associated with `launchurl` selecting a browser, switch activities etc.  These things ought to be simple.)


0-8-15:
Please see `README.md` section *Issues* for things left to be done.  Suggestions and recommendations solicited.

mgorges:
@0-8-15 this sounds really interested but I am not sure if I am the best person to review it. That said it seems to be self-contained so merging should not be a large risk? - Some of it would go into modules at a later point?

0-8-15:
On Thu, 14 Nov 2019 21:53:44 -0800
Matthias Görges <notifications@github.com> wrote:

> @0-8-15 this sounds really interested but I am not sure if I am the
> best person to review it. That said it seems to be self-contained so
> merging should not be a large risk?

Being sure that it can't do any damage at all was the point to first
come up with self-contained app in the demo directory.

> - Some of it would go into modules at a later point?

That's the idea.

Doing so will likely/certainly require changes to be able to install
`.jar`'s from modules.  That would be a (minor) change to the base
system, which should be in a PR of it's own.

Furtheremore I'd personally like to first get Java class field
accessors done.  And maybe a few more examples.  (e.g., the foreground
service should be a no-brainer this way.)

However until this is done, distributing the demo might be inspirative
and solicity comments.

PR #281, opened 2020-04-01
UIFORM: adds function to clear required field

karliwalti:
function to clear required flags in store (i.e. when a reset is needed after a cancel)


ddunsmuir:
This works for me, but please update the comment before the function, it looks like it's a copy of the uiform-required-set comment, and also remove the unnecessary changes to the brackets much further down the file.

karliwalti:
Ok, that should do it

PR #282, opened 2020-04-01
[NEW][ROTATION] add the android module for determining phone rotation

karliwalti:
this module adds IMU process rotation data as sensor module. Only works for Android atm. 
Would be great if someone can expand with iOS version


mgorges:
How are these different from the yaw, pitch, and roll in the [gyro](https://github.com/part-cw/lambdanative/wiki/Index-of-Module-gyro) module, and why not integrate them there if they are related?

karliwalti:
See https://developer.android.com/guide/topics/sensors/sensors_motion

gyro is like accelerometer raw sensor data (from hardware) and is a rate. rotation is a position vector (IMU calculation) integrated from the sensors mentioned before.

karliwalti:
For IOS, I would say the quaternion is the closest type to this https://developer.apple.com/documentation/coremotion/cmquaternion

mgorges:
Added iOS code in d9d27a6 - also please document this one in the Wiki. Thanks!

PR #283, opened 2020-04-01
[NEW][MAGNETOMETER] add compas reading to modules for android

karliwalti:
Adds compass sensor readings (only Android available, if someone with iOS testing possibilities cann add those too?)

demo app for all sensors will follow...


mgorges:
iOS might have information at [Getting Heading and Course Information](https://developer.apple.com/documentation/corelocation/getting_heading_and_course_information) - there is a reference to `CLLocationManager` in [loaders/ios/mo_location.mm](https://github.com/part-cw/lambdanative/blob/master/loaders/ios/mo_location.mm)

karliwalti:
I would have said this: https://developer.apple.com/documentation/coremotion/cmmagnetometerdata

mgorges:
I guess the title is misleading then as it mentions compass, but what you actually want is the magnetic field information.

karliwalti:
true yes, this is why I called the module magnetometer  :-)

mgorges:
Merged, with iOS code added in de77329. Please add documentation to the [Wiki](https://github.com/part-cw/lambdanative/wiki)

PR #284, opened 2020-04-11
Covid-19

gassechen:
Hi. I would like to make an inquiry. What benefit would generate an alkaline environment by spraying an apple vinager water solution continuously in a confined space. Covid-19
Thank you and sorry for the means of communication.


mgorges:
There is a reason for this branch staying at an earlier version of Gambit, which is to allow reproducible software building. Also this is not the right place to ask for health advice - sorry.

PR #293, opened 2020-07-14
GAMBIT: Fix IPv6 support

0-8-15:


mgorges:
Does this work for all platforms?

Issue #234, opened 2019-05-05
keyboard signs in Uniforms broken on Mac


karliwalti:
**Issue:** The UIform keyboards do not display the special function pictures anymore on Mac. Instead there is a white box
**Expected behaviour:** full keyboard with "back" or "enter" sign
**Reproduce:** build DemoUIForms on mac (10.14.3) with latest lambdanative, go to the widget page, select a form field

mgorges:
We have encountered the problem as well - it is also present in some iOS cases and has something to do with low-level OpenGL handling of vertices and transparency _(in glcore.scm)_ which Apple changed. We have not yet been able to identify a solution.

mgorges:
I have created an app to illustrate the problem with clipping? [OpenGLTesting](https://github.com/mgorges/lambdanative-testing/tree/master/apps/OpenGLTesting) - It seems that those textures rendered at runtime are fine, while those shown later are not?

mgorges:
I have narrowed the problem down to a problem with `glGenTextures` in  [ln_glcore/glcore.scm#L395](https://github.com/part-cw/lambdanative/blob/master/modules/ln_glcore/glcore.scm#L395) which sets `u32t=0`, instead of an increasing integer, when un-hiding textures. No idea yet how to solve it though the `glGetError` is empty; potentially something about the OpenGL Context not being active within that thread?

peterlew:
I'm also concerned about it being a thread issue, but haven't found anything to suggest that the OpenGL Context is inactive or different in any way when the later draw is called.

 I tried calling `glGenTextures` for all textures on the first draw, and all the textures were properly assigned increasing `u32t ` ids. Unfortunately, this did not solve the problem. Maybe the unexpected behavior of `glGenTextures` is a side effect instead of a cause of the problem.

peterlew:
It seems that the OpenGL Context could be inactive, and the texture would be marked as initialized anyway. This should be fixed by the commit above.

PR #297, opened 2020-07-31
Fix of Issue #234

peterlew:
[Issue #234](https://github.com/part-cw/lambdanative/issues/234). 

The issue is that `_glCoreTextureCreate` sometimes does not properly set up the texture, but marks it as initialized anyway. This fix adds a further check on the proper operation of the GL Context, creates an error, and lets the texture be re-initialized on next call. I believe this stems from being called in a different thread from the main GL Context. 

This also fixes what I believe to be a typo in `_glCoreTextureBind`, and adds helpful functions `glGetError` and `glIsEnabled` to glcore-ffi.scm

### DIFF
--- a/modules/ln_glcore/glcore-ffi.scm
+++ b/modules/ln_glcore/glcore-ffi.scm
@@ -161,6 +161,10 @@ ___result = GL_CLAMP_TO_EDGE;
 
 (define glPopMatrix (c-lambda () void "glPopMatrix"))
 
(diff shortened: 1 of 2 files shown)


mgorges:
Well done @peterlew - the key piece is the check for `(= (glIsEnabled GL_TEXTURE_2D) 0))`, which seems to resolve it for me. The piece in `_glCoreTextureBind` is not necessary, but probably good to have.

PR #302, opened 2020-08-17
BUILD: Add SYS_CXX and CXX subsitution

0-8-15:


mgorges:
Sorry I see what it does but why do we need this? (And why is this selective for some platforms, e.g. not for Android on Linux?)

0-8-15:
Am Mon, 17 Aug 2020 15:30:12 -0700

> Sorry I see what it does but why do we need this? (And why is this

I need to build/link a library written in C++.

> selective for some platforms, e.g. not for Android on Linux?)

So far it's only tested for Android cross developed from Linux.
(Though later I found Android API>=21 will need additional work if the
STL is used by the library.)

For other platforms I only added the code when it was obvious to me.
(I.e., I don't have access to an Apple environment and don't know QNX's
C++ compiler.)

mgorges:
This does **not** cover Android on Linux, there is only a change to `targets/android/_host_macosx`. I don't want to go and test all those other platforms either, but it seems strange to have it applied selectively to only some platforms? Also how do we test this?

mgorges:
450a800 adds a few more platforms but this still doesn't have 100% coverage!

mgorges:
> I need to build/link a library written in C++.
To follow up on this can you provide a test case, also I thought that OpenSSL, which we build, is written in C++ already and it has always built just fine?

PR #308, opened 2020-09-07
extending rotation functions

karliwalti:
This increases the utility of the rotation vectors (quarternions)


mgorges:
Please add documentation for this to the Wiki at: [Index of Module rotation](https://github.com/part-cw/lambdanative/wiki/Index-of-Module-rotation), as I will not do this for you. Also, the function naming convention is inconsistent, as not all functions start with `rotation`

PR #310, opened 2020-09-09
updates the rotation functions to naming convention

karliwalti:
all functions start now with `rotation-` as requested

matches the wiki entries previously pushed.


mgorges:
Thanks - I think for `rotation-orientation` `rotation-get-orientation` would also be totally acceptable as a function name. I also don't think `rotation:deg2rad` is used, correct?

karliwalti:
not used, but useful maybe in future.

PR #313, opened 2020-09-10
REDCAP: optimisation and new function

karliwalti:
Sorry for again having more history in it than necessary. Each time I do a rebase there is new commits to apear. Hope improve soon

### DIFF
--- a/modules/redcap/redcap.scm
+++ b/modules/redcap/redcap.scm
@@ -691,7 +691,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
    ;; Check if we have a valid connection before proceeding
     (if (and filesize (fx= (httpsclient-open host redcap:port) 1))
       (let* ((fh (open-input-file filename))
-             (buflen 100000)
+             (buflen (if (or (string=? (system-platform) "android") (string=? (system-platform) "ios")) 50000 (if (string=? (system-platform) "linux") 100000 100000 )))
              (buf (make-u8vector buflen)))
         (httpsclient-send request-vector)
         (let loop ((start 0) (end (if (fx< filesize buflen) filesize buflen)))
@@ -704,6 +704,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
                 (httpsclient-send (subu8vector buf 0 len))
                 (httpsclient-send buf)
               )
+              (thread-sleep! 0.0001) ;;allow GUI to refresh
               (loop (fx+ start buflen) (if (fx> (fx+ start (fx* 2 buflen)) filesize) filesize (fx+ end buflen)))
             )
           )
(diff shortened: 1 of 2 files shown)


modules/redcap/redcap.scm
```diff
@@ -691,7 +691,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
    ;; Check if we have a valid connection before proceeding
     (if (and filesize (fx= (httpsclient-open host redcap:port) 1))
       (let* ((fh (open-input-file filename))
-             (buflen 100000)
```
ddunsmuir, line 649: Linux gives no change....either 100000 or 100000...?? So this could be removed?

karliwalti:
there might be needs to change on embedded versions in the future.

ddunsmuir:
Your UIFORM commits basically do nothing. They do not do what the commit message says, so yes as the pull request says, this is just changes to REDCap. The last commit is not a change to "file-export", it is a change to "file-import". We would appreciate if this was re-pushed with all this fixed so our history makes sense and the commit messages match what actually happened. I will test out this project export. Thanks!

mgorges:
@ddunsmuir I always fix the commit message when I squash and merge the commit; that way we only have a single commit in the history and not 4. As for uniform, I saw that too, but it all it does is add a blank line, which isn't great but no harm. Thanks for testing the export.

karliwalti:
mea culpa, the file upload to redcap is called file import not export.

ddunsmuir:
@karliwalti the optimisation to the file-export works great.
The new redcap-export-project does not work. Did you test this? It at least does not work with REDCap 9.5.33. The reason is that the json starts with "{" and ends with "}" whereas the json from a record export for example starts with "[{" and ends with "}]" regardless of whether there are 1 or more results. I guess this lack of square brackets is because for the project export there will only ever be one result. What this all means is that the redcap:jsonstr->list fails to process the output.
- Please update that procedure to check for both possible string ends or accept an optional flag to deal with this.
- Please rename this procedure redcap-export-project-info. "Export Project Info" is the API Method (at least according to the API Playground). All this method does is export the project title and other metadata, nothing major, so I think the revised name is better.
- Please add a space on your two lines with the error, so it's "REDCap error: Incomplete json " and the output is not right up against it.
Thanks!

karliwalti:
We are running redcap 10.1, I can't test with your version.

I am using this function in over 10 projects for quite a while now and did not encounter any issues yet

It returns

```project_id, project_title, creation_time, production_time, in_production, project_language, purpose, purpose_other, project_notes, custom_record_label, secondary_unique_field, is_longitudinal, has_repeating_instruments_or_events, surveys_enabled, scheduling_enabled, record_autonumbering_enabled, randomization_enabled, ddp_enabled, project_irb_number, project_grant_number, project_pi_firstname, project_pi_lastname, display_today_now_button, missing_data_codes, external_modules```

I will change name and what you suggest

ddunsmuir:
Ah I see. We are getting an upgrade to 10.0 in a couple weeks, but we may still work with collaborators that use an older REDCap, so it would be good to account for that.

mgorges:
Closing, superseded by #313

PR #320, opened 2020-09-25
GAMBIT: fix ipv6 support2

0-8-15:
This PR extends the existing patch to gambit, which fixed IPv6 support on server side.  Adding client side fixes.

As this changes a pre-compiled `.scm` file it adds a second patch changing the latter.  (This second patch is huge and completely boring as it can be re-created from the first change.  However this would require to go through the boostrap process of gambit.  IMHO it's better to include another patch and NOT change the build process.)


libraries/libgambit/make.sh
```diff
@@ -48,6 +48,7 @@ package_configure $EXTRACONF
 if [ "$SYS_PLATFORM" = "$SYS_HOSTPLATFORM" ]; then
   package_make
 else
+  echo 'exit 0' > gsc-boot && chmod +x gsc-boot
```
mgorges, line 51: Would you mind explaining to me what this does? I can't quite figure this out, sorry.

libraries/libgambit/fix_IPv6_precompiled.patch
```diff
@@ -0,0 +1,5220 @@
...
+                ___STR4(46,115,99,109)
+ ___DEF_SUB_STR(___X1,4UL)
+@@ -8573,19 +8573,23 @@
+                ___VEC0
+ ___DEF_SUB_U8VEC(___X125,4UL)
+                ___U8VEC4(0x7fL,0x0L,0x0L,0x1L)
+ ___DEF_SUB_STR(___X126,1UL)
+                ___STR1(42)
+-___DEF_SUB_STRUCTURE(___X127,6UL)
++___DEF_SUB_U16VEC(___X127,8UL)
```
mgorges, line 94: This is really the only change, everything else is renumbering things, right?

mgorges (commented):
Thanks for looking into this - I am sorry but would you mind explaining two of the changes to make sure I better understand what I merge, please? Also, how would I go about testing this?

0-8-15:
Sure:

Am Fri, 25 Sep 2020 09:35:36 -0700

> 
> Thanks for looking into this - I am sorry but would you mind
> explaining two of the changes to make sure I better understand what I
> merge, please? Also, how would I go about testing this?
> 
> > @@ -48,6 +48,7 @@ package_configure $EXTRACONF  
>  if [ "$SYS_PLATFORM" = "$SYS_HOSTPLATFORM" ]; then
>    package_make
>  else
> +  echo 'exit 0' > gsc-boot && chmod +x gsc-boot

Looking into it I see that this made it into the wrong of the two
patches.  Sorry.

The point of this line is to avoid issues in case the compile/make
finds the source code change (by the other patch) to be newer than the
change to the pre-compiled C file.  If so it would to what the gambit
bootstrap process would do: invoke gsc-boot to re-compile the .scm into
the corresponding .c file.

This will fail if gsc-boot does not exist and thus break the build.

What I did here was to copy what the "normal" gambit build does: create
a fake/temporary/bootstrap `gsc-boot` which does not change the .c
file.  Then the gambit build will continue as expected.

> Would you mind explaining to me what this does? I can't quite figure
> this out, sorry.

No problem.  It's weird.  As bootstrapping these things is.

> > ++,___DEF_CNS(___REF_SUB(244),___REF_SUB(245))  
> ++,___DEF_CNS(___REF_SUB(246),___REF_SUB(247))
> ++,___DEF_CNS(___REF_SUB(248),___REF_SUB(249))
> + ___END_CNS
> + 
> + ___DEF_SUB_STR(___X0,4UL)
> +                ___STR4(46,115,99,109)
> + ___DEF_SUB_STR(___X1,4UL)
> +@@ -8573,19 +8573,23 @@
> +                ___VEC0
> + ___DEF_SUB_U8VEC(___X125,4UL)
> +                ___U8VEC4(0x7fL,0x0L,0x0L,0x1L)
> + ___DEF_SUB_STR(___X126,1UL)
> +                ___STR1(42)
> +-___DEF_SUB_STRUCTURE(___X127,6UL)
> ++___DEF_SUB_U16VEC(___X127,8UL)
> 
> This is really the only change, everything else is renumbering
> things, right?

Probably yes.

There is actually only a single line changed in the _io.scm file.  This
should not add too much code.  I.e. this looks roughly like the
resulting consequence.

The second consequence is that gambit will renumber so many things it
finds now.

Frankly: I trusted gambit wrt. the outcome.  I built gambit with the
source change only.  Then tested the result.  Then committed the first
patch.  Only then I decided that it would be crazy to modify the gambit
build in lambdanative to go though a needless "make; make bootstrap;
make clean; make" cycle -- only in order to re-do what the distribution
included already: a pre-compile .c from the .scm file.

PR #321, opened 2020-09-25
SCM: add target specific directory definition to gambit invocation

0-8-15:
Gambit allows to refer to various installation dependent directories
using the ~~NAME symtax - e.g., as in `(include "~~lib/my/file.scm")`.
However gambit is not aware of cross-compiling.  So `~~lib` will refer
to the hosts version.

This change adds another option to the gambit invocation, which allows
to distinguish between the host and the target specific files of
gambit.  At least for the library.

The use case this is required for is not (yet) publically available.
It involves a library wich includes gambit source code using
cond-expand to include use cases like file locking for instance -
which is a horror to provide in a platform agnostic way.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -121,7 +121,7 @@ compile_payload_scm()
 	if [ -f $scm_hdr ]; then scm_hdr="-e '(load \"$scm_hdr\")'"; else scm_hdr=""; fi
 	gsc_processing=""
 	# if [ $SYS_VERBOSE ]; then gsc_processing="$gsc_processing -expansion"; fi
-        veval "$SYS_GSC -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
+        veval "$SYS_GSC -:~~tgtlib=${SYS_PREFIX}/lib -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
         if [ $veval_result != "0" ]; then rmifexists "$scm_ctgt"; fi
         assertfile "$scm_ctgt"
         rmifexists "$scm_otgt"
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -121,7 +121,7 @@ compile_payload_scm()
 	if [ -f $scm_hdr ]; then scm_hdr="-e '(load \"$scm_hdr\")'"; else scm_hdr=""; fi
 	gsc_processing=""
 	# if [ $SYS_VERBOSE ]; then gsc_processing="$gsc_processing -expansion"; fi
-        veval "$SYS_GSC -:~~tgtlib=${SYS_PREFIX}/lib -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
+        veval "$SYS_GSC -:~~tgt=${SYS_PREFIX} -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
         if [ $veval_result != "0" ]; then rmifexists "$scm_ctgt"; fi
         assertfile "$scm_ctgt"
         rmifexists "$scm_otgt"


0-8-15:
As I submitted the PR it ocured to me that maybe we rather had `-:~~tgt=${SYS_PREFIX}` defined instead of `-:~~tgtlib=${SYS_PREFIX}/lib`.  Would be more general.  (Just untested, but should work.)

languages/scm.sh
```diff
@@ -121,7 +121,7 @@ compile_payload_scm()
 	if [ -f $scm_hdr ]; then scm_hdr="-e '(load \"$scm_hdr\")'"; else scm_hdr=""; fi
 	gsc_processing=""
 	# if [ $SYS_VERBOSE ]; then gsc_processing="$gsc_processing -expansion"; fi
-        veval "$SYS_GSC -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
+        veval "$SYS_GSC -:~~tgtlib=${SYS_PREFIX}/lib -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
```
mgorges, line 124: `tgtlib` is a new variable you are introducing, right as it doesn't overwrite any internal ones I can see in the gambit source?
Note to self: The syntax for this is documented here: http://www.iro.umontreal.ca/~gambit/doc/gambit.html#Runtime-options

mgorges:
I highly doubt our team will ever use any `cond-expand` in our code, as all occurrences in the framework where such a construct is used were contributed by you or another library we included, but I don't see any harm in adding this. No preference to use of `~~tgt` or `~~tgtlib` so if you feel there is the need to get access to the larger folder then please amend the pull request otherwise I will merge it once you confirmed it is a new variable, not overwriting an existing one.

0-8-15:
Am Fri, 25 Sep 2020 11:04:55 -0700

> I highly doubt our team will ever use any `cond-expand` in our code,
> as all occurrences in the framework where such a construct is used
> were contributed by you or another library we included, but I don't
> see any harm in adding this. No preference to use of `~~tgt` or
> `~~tgtlib` so if you feel there is the need to get access to the
> larger folder then please amend the pull request otherwise I will
> merge it once you confirmed it is a new variable, not overwriting an
> existing one.

OK.  I shall amend it.  (Though not today anymore.)

BTW: Once you learn about the reason `cond-expand` is also `SRFI-0` I
bet your team will change its mind.  ;-)

0-8-15:
Am Fri, 25 Sep 2020 10:58:26 -0700

> 
> 
> 
> > @@ -121,7 +121,7 @@ compile_payload_scm()  
>  	if [ -f $scm_hdr ]; then scm_hdr="-e '(load \"$scm_hdr\")'";
> else scm_hdr=""; fi gsc_processing=""
>  	# if [ $SYS_VERBOSE ]; then gsc_processing="$gsc_processing
> -expansion"; fi
> -        veval "$SYS_GSC -prelude \"$scm_opts\" -c -o $scm_ctgt
> $gsc_processing $scm_hdr $scm_src"
> +        veval "$SYS_GSC -:~~tgtlib=${SYS_PREFIX}/lib -prelude
> \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
> 
> `tgtlib` is a new variable you are introducing, right as it doesn't
> overwrite any internal ones I can see in the gambit source? Note to
> self: The syntax for this is documented here:
> http://www.iro.umontreal.ca/~gambit/doc/gambit.html#Runtime-options

True: it's on purpose not in conflict with gambit itself nor in
conflict with lambdanative.  The name is even unlikely to be ever used
by gambit itself.  (Otherwise it had been `target`.)

The intention is to add a cross-compile option to gambit code within
lambdanative.  Independent of possible future extensions to gambit.

Gambit itself is actually target agnostic.  (Which made me - originally
- question why the .scm source is compiled anew for each target.
According to gambit the result is supposed to be identical.)

However since lambdanative does actually provide multiple targets which
- in some cases - do need target specific code - it is great to be able
to do this at the Scheme level -- including all the features (not yet
used by lambdanative but right at the finger tips and at least use by
me ;-) like conditionally including other files).

The possible alternative to cond-expand would be to always do it using
#ifdef hacks in the resulting C code is much, much more cumbersome and
error prone.  So I welcome cond-expand very much.

mgorges:
> Gambit itself is actually target agnostic. (Which made me - originally - question why the .scm source is compiled anew for each target. According to gambit the result is supposed to be identical.)

I assume we could explore caching the build/*.c files globally, but I am afraid I get yelled at again if I brake things doing so, so not a priority for the time being.

PR #322, opened 2020-09-25
LIBPCAP: change build

0-8-15:
This modification likely requires review!  I just tried to use it,
found not working and changed what was required in my case.

No knowledge here what could break.


mgorges:
This breaks on `macos` where _Include_ and _include_ are the same folder and now they appear in the `win32` folder directly.

PR #325, opened 2020-09-25
srcipts/package: enable rename of download

0-8-15:


scripts/package.sh
```diff
@@ -140,7 +140,11 @@ package_download_ball()
 {
   pkg_url=$1
   pkg_hash=$2
-  pkg=$SYS_PREFIXROOT/packages/`basename $pkg_url`
+  pkg_basename=$3
+  if [ "$pkg_basename" = "" ]; then
```
mgorges, line 144: While this works, I never know if it is better to test with `-z` or use the `"X$2" = "X"` approach we use a lot

0-8-15:
Am Fri, 25 Sep 2020 12:26:43 -0700

> 
> 
> 
> > @@ -140,7 +140,11 @@ package_download_ball()  
>  {
>    pkg_url=$1
>    pkg_hash=$2
> -  pkg=$SYS_PREFIXROOT/packages/`basename $pkg_url`
> +  pkg_basename=$3
> +  if [ "$pkg_basename" = "" ]; then
> 
> While this works, I never know if it is better to test with `-z` or
> use the `"X$2" = "X"` approach we use a lot


Neither do I.

Both ought to be valid shell code.

The additional capital X within double quoted strings looks to the
graybeard like an overkill applying two strategies known since the dawn
of sh at the same time: X$VAR = X vs. "$VAR" = "" - not really the way
I'd write readable code.  Hence I tend to prefer the quotes-version as
it is more in-line with other languages.  The -z strategy is just yet
another one... Uhm.  Did I ever mention that I have a reason to not
like sh for it being too close to Perl in having too many ways to do
the same thing?  And, additionally - and worse - for re-parsing it's
input several times?  Nah, no flame war.  sh is there.  But I don't
discuss these things.

On a related note: I'd really like to suggest a policy to NOT mix
taste/style-related changes like adding (likely irrelevant) double
quotes in the very same change as actually functional modifications.
It makes it tedious to figure out what was done.

PR #329, opened 2020-09-26
GLGUI: chat - added comments and minor runtime optimizations

0-8-15:
Additionally I had to re-format the source according to widespread
conventions for readbility and used keyword like `when` and `unless`
to make the source more compact and easier to digest.

Sorry for the latter.

### DIFF
--- a/modules/ln_glgui/chat.scm
+++ b/modules/ln_glgui/chat.scm
@@ -204,54 +204,59 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ))
 
 (define (glgui:chat-input g wgt type mx my)
-  (let* ((x (fix (glgui-widget-get-dyn g wgt 'x)))
+  (let* ((mx-fixed (fix mx)) ;; Q: Can `mx` or `my` ever not be fixnums?
+         (my-fixed (fix my)) ;; converting them once here for both readability and speed
+         (x (fix (glgui-widget-get-dyn g wgt 'x)))
          (y (fix (glgui-widget-get-dyn g wgt 'y)))
          (w (fix (glgui-widget-get-dyn g wgt 'w)))
          (h (fix (glgui-widget-get-dyn g wgt 'h)))
          (cb (glgui-widget-get g wgt 'callback))
          (dh (glgui-widget-get g wgt 'dh))
          (lst (glgui-widget-get g wgt 'list))
-         (n (length lst))
+         (n (length lst)) ;; length is O(n) - try to avoid
          (val (glgui-widget-get g wgt 'offset))
          (valmin 0)
          (valmax_temp (glgui-widget-get g wgt 'valmax))
          (valmax (if valmax_temp valmax_temp n))
-         (old (glgui-widget-get g wgt 'old))
-         (drag (glgui-widget-get g wgt 'drag))
-         (oldy (glgui-widget-get g wgt 'oldy))
-         (fsty (glgui-widget-get g wgt 'fsty))
+         (old (glgui-widget-get g wgt 'old)) ;; boolean: "touched/clicked"
+         (drag (glgui-widget-get g wgt 'drag)) ;; boolean: "touched/clicked" and "motion seen"
+         (oldy (glgui-widget-get g wgt 'oldy)) ;; y coordinate upon "touched/clicked" start
+         (fsty (glgui-widget-get g wgt 'fsty)) ;; y coordinate upon "touched/clicked" start
          (space (/ (modulo h dh) 2.))
-         (inside (and (fx> (fix mx) x) (fx< (fix mx) (fx+ x w)) (fx> (fix my) y) (fx< (fix my) (fx+ y h)))))
-     (cond
-       ((and (fx= type EVENT_MOTION) old) ;; drag
-         (if (> (abs (- my fsty)) 5) (begin
-           (glgui-widget-set! g wgt 'offset (max valmin (min valmax
-              (+ val (* (- valmax valmin) (/ (flo (* 1. (- oldy my))) (flo h)))))))
-           (glgui-widget-set! g wgt 'oldy (fix my))
-           (glgui-widget-set! g wgt 'drag #t)
-        ))
-       )
-       ((and inside (fx= type EVENT_BUTTON1DOWN)) ;; touch down
-         (glgui-widget-set! g wgt 'oldy (fix my))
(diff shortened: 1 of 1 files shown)


mgorges:
I have to disagree - there is not a single place in the entire codebase, where `unless` or `when` are used, as such this might confuse instead of help. You said not to make personal style related changed in one of your last pull request, but to me this clearly is, as this is your preference, not what the project has as a whole.

mgorges:
To make sure this is documented. The expression equivalence for `when` is:

``` 
(when <test>
  <expression1>
  <expression2>
)
```
is equivalent to
```
(if <test>
  (begin
    <expression1>
    <expression2> 
  )
)
```

mgorges:
The other case for `unless` includes an additional `not`, so:
```
(unless <test>
  <expression1>
  <expression2>
)
```
is equivalent to 
```
(if (not <test>)
  (begin
    <expression1>
    <expression2> 
  )
) 
```

modules/ln_glgui/chat.scm
```diff
@@ -204,54 +204,59 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-))
+         (inside (and (fx> mx-fixed x) (fx< mx-fixed (fx+ x w)) (fx> my-fixed y) (fx< my-fixed (fx+ y h)))))
+    (cond
+     ((and (fx= type EVENT_MOTION) old) ;; drag
+      (if (> (abs (- my fsty)) 5)
+          (let* ((calculated (+ val (* (- valmax valmin) (/ (flo (* 1. (- oldy my))) (flo h))))) ;; Why these `flo`?
+                 (clipped (max valmin (min valmax calculated))))
+            (glgui-widget-set! g wgt 'offset clipped)
+            (glgui-widget-set! g wgt 'oldy my-fixed)
+            (unless drag (glgui-widget-set! g wgt 'drag #t)))))
```
mgorges, line 234: This piece is new, it tests if `drag` is already set.

modules/ln_glgui/chat.scm
```diff
@@ -204,54 +204,59 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+     ((and inside (fx= type EVENT_BUTTON1DOWN)) ;; touch down
+      (glgui-widget-set! g wgt 'oldy my-fixed)
+      (glgui-widget-set! g wgt 'fsty my-fixed)
+      (glgui-widget-set! g wgt 'old #t)
+      (glgui-widget-set! g wgt 'drag #f))
+     ((and old (fx= type EVENT_BUTTON1UP)) ;; touch release
+      (cond
+       ((not inside) ;; cancel drag
+        (if old (glgui-widget-set! g wgt 'old #f))
+        (if drag (glgui-widget-set! g wgt 'drag #f)))
```
mgorges, line 244: this piece is new too.

modules/ln_glgui/chat.scm
```diff
@@ -204,54 +204,59 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+       (else
+        (glgui-widget-set! g wgt 'offset (round val)) ;; snap to place?
+        (if (and (not drag) inside cb)
+            (let ((cur (round (+ val (/ (- (+ y h) (+ my space)) dh) -.5))))
+              (when (and (>= cur 0) (>= cur val) (< cur n))
+                (glgui-widget-set! g wgt 'current cur)
+                (if (procedure? cb) (cb g wgt type mx my)))))
+        (if old (glgui-widget-set! g wgt 'old #f))
```
mgorges, line 252: new test added

modules/ln_glgui/chat.scm
```diff
@@ -204,54 +204,59 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+        (if drag (glgui-widget-set! g wgt 'drag #f)))))
+     (else ;; neither button1 nor motion
+      (unless inside
```
mgorges, line 255: this piece is also new

modules/ln_glgui/chat.scm
```diff
@@ -204,54 +204,59 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ))
 
 (define (glgui:chat-input g wgt type mx my)
-  (let* ((x (fix (glgui-widget-get-dyn g wgt 'x)))
+  (let* ((mx-fixed (fix mx)) ;; Q: Can `mx` or `my` ever not be fixnums?
```
mgorges, line 207: I don't know, but we use `fx` math later, which segfaults if it is not without pointing to where the error is, which is extremely annoying to debug.

0-8-15:
Am Sat, 26 Sep 2020 13:02:00 -0700

> I have to disagree - there is not a single place in the entire
> codebase, where `unless` or `when` are used, as such this might
> confuse instead of help. You said not to make personal style related
> changed in one of your last pull request, but to me this clearly is,
> as this is your preference, not what the project has as a whole.

Sorry for that.  --  As I wrote in the commit message.

Initially I just wanted to understand the code (since I was hunting for
the bug I made elsewhere).

Then (with the background of ~30 years of Scheme experience) I felt that
the unusual formatting was the part holding me back.  There is a rather
long tradition to use automatic code formatting among most Schemers.
Essentially the editors do this for you.  A few more rules are usually
recommended, e.g., multiple let-bindings on a line are OK if the column
76 is not crossed, but for let* and letrec this is never done; always
have a space between adjacent closing and opening parenthesis (since
some reader extensions like SRFI-110 or gambits 'six' might convert
these things); nothing behind `begin` if the block spans multiple lines.

Additionally working with `git` (which I don't really like for a
version control system, but it won the world) taught me the habit to
let the editor automatically remove trailing white space.

Then I ran into the repeated use of `length` for the same input.
`length` is O(n) and thus usually avoided for performance.  As we are
in the many-time-per-second part of the code I just removed that one.
Same for the repeated use of `fix` on the same input.  As I now was in
the optimized-a-few-thing-here-mode I added those "don't modify a hash
table for no reason" checks.  For consistency I added another case in
the else branch.

At that point I was convinced that I should now understand the code and
can't find a bug there.  Hence I looked an found it elsewhere.

What to do with the change now?  Throw away the improvements just
because they contain what I would call a code cleanup because that
would include personal preferences?  Or propose it as a PR and have you
guy decide?

I opted for the latter.

Issue #202, opened 2018-07-28
glgui: high idle CPU load


lighhw68:
Compiled a few of the demo apps and noticed that on Linux I get easily 20% idle cpu load. 

I am worried how that would affect battery life on phones or laptops? 

Is that fixable somehow?

mgorges:
The demo apps are purposefully kept simple; you need to sleep more (how much depends on your app) to reduce CPU load. E.g. add something like this to your event loop (the one with ` (lambda (t x y)`:

    ;; If all we do is redraw sleep a bit more 40usec (10 was from normal sleep)
    (if (fx= t EVENT_REDRAW)
      (thread-sleep! 0.04)
    )

lighhw68:
Looks like it should be good enough .. don't want to turn on my big computer until the worst of the heat wave is over.

Is the periodic refresh/redraw needed for some glgui-internal reason or  because X11 expose and similar events are not handled?

mgorges:
You need to event loop to keep the screen refreshed (e.g. when moved), deal with input events, and update values as they change. Let us know once you tested this if this works for you in which case the ticket can be closed.

lighhw68:
Tested with Calculator app the idle load on Linux goes from 23% down to 8%. Much better but not quite sure this is good enough. 

Can't the EVENT_REDRAW get generated only when needed? In principle it should be possible to detect conditions that require a redraw?

ionathanch:
According to https://tronche.com/gui/x/xlib/events/exposure/expose.html#XExposeEvent, `XExposeEvent` has a `count` field for how many more `XExposeEvent`s are to follow, so we could check for `count == 0` to trigger a redraw. Would this patch help things?

```diff
diff --git a/loaders/x11/x11_microgl.c b/loaders/x11/x11_microgl.c
index 947ce44..082da1a 100644
--- a/loaders/x11/x11_microgl.c
+++ b/loaders/x11/x11_microgl.c
@@ -218,7 +218,9 @@ void microgl_pollevents(void)
         motion=1;
         break;
       case Expose:
-        expose=1;
+        if (event.count == 0) {
+          expose=1;
+        }
         break;
       case ClientMessage:  
         if( (Atom) event.xclient.data.l[ 0 ] == win.WMDeleteWindow )
```

lighhw68:
Doesn't work, I get

```
x11_microgl.c: In function ‘microgl_pollevents’:
[...]

PR #330, opened 2020-09-26
GLGUI: reduce idle load

0-8-15:
Note: When other threads notice that the GUI thread should stop
waiting and thus reducing responsiveness : call `(glgui-wakeup!)` .

### DIFF
--- a/modules/ln_glgui/glgui.scm
+++ b/modules/ln_glgui/glgui.scm
@@ -230,12 +230,43 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ;; process an input event
 ;; 20100519: allow multiple guis
 ;; 20100804: support gui offset
-(define (glgui-event guis t x0 y0)
-  (if (and glgui:active app:width app:height)
-    (let ((gs (if (list? guis) guis (list guis))))
-      (if (fx= t EVENT_REDRAW) (apply glgui:render gs)
-        (apply glgui:inputloop (append (list t x0 y0) gs)))
-   )))
+
+(define glgui-frame-period-max (make-parameter 0.5)) ;; How long to sleep at most in redraw.
+
+;; glgui-wakeup! - a thunk, which when called will immediately unblock
+;; the thread waiting in glgui-event.  Should be called if other
+;; threads notice the event loop should proceed.
+(define glgui-wakeup!)
+
+(define glgui-event
+  (let ((consecutive-redraw-count 1)
+        (step 0.05)
+        (wait-mutex (make-mutex 'glgui-event))
+        (wait-cv (make-condition-variable 'glgui-event)))
+    (define (wakeup!)
+      (condition-variable-signal! wait-cv))
+    (define (reset-wait!)
+      (set! consecutive-redraw-count 1))
+    (define (wait-for-time-or-signal!)
+      (if (let ((moment (min (glgui-frame-period-max) (* consecutive-redraw-count step))))
+              (mutex-unlock! wait-mutex wait-cv moment))
+          (reset-wait!)
+          (set! consecutive-redraw-count (fx+ consecutive-redraw-count 1))))
+    (define (glgui-event guis t x0 y0)
+      (if (and glgui:active app:width app:height)
+          (let ((gs (if (list? guis) guis (list guis))))
+            (if (fx= t EVENT_REDRAW)
+                (when (mutex-lock! wait-mutex 0)
+                  (apply glgui:render gs)
+                  (wait-for-time-or-signal!))
+                (begin
+                  (reset-wait!)
+                  (apply glgui:inputloop (append (list t x0 y0) gs)))))
+          (if (fx= t EVENT_REDRAW)
+              (wait-for-time-or-signal!)
+              (reset-wait!))))
+    (set! glgui-wakeup! wakeup!)
+    glgui-event))
 
 ;; provide a screen shot
 (define (glgui-screenshot)
--- a/modules/ln_glgui/glgui.scm
+++ b/modules/ln_glgui/glgui.scm
@@ -316,9 +316,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
                   (apply glgui:inputloop (append (list t x0 y0) gs)))))
           (if (fx= t EVENT_REDRAW)
               (wait-for-time-or-signal!)
-              (begin
-                (thread-sleep! step)
-                (reset-wait!)))))
+              (if frame-period-custom
+                  (thread-sleep! (frame-period-custom 1))
+                  (begin
+                    (thread-sleep! step)
+                    (reset-wait!))))))
     (set! glgui-wakeup! wakeup!)
     (set! glgui-timings-set! timings-set!)
     glgui-event))
--- a/modules/ln_glgui/glgui.scm
+++ b/modules/ln_glgui/glgui.scm
@@ -329,7 +329,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   (define (wait-for-sec _) (seconds->time (+ ##now sec)))
   (define (no-wait _) 0)
   (cond-expand
-   ((or android iod)
+   ((or android ios)
     ;; TBD: convey the time value to signaling code.
     ;; switch delays to zero
     (glgui-timings-set! frame-period-custom: no-wait))
@@ -338,7 +338,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 (define (glgui-timings-at-10msec!) (glgui-timings-at-sec! 0.01))
 
 (cond-expand
- ((or android iod)
+ ((or android ios)
   ;; 50 Hz should be enough - TBD: not yet effective
   (glgui-timings-at-sec! 0.02))
  (else))
--- a/modules/ln_glgui/glgui.scm
+++ b/modules/ln_glgui/glgui.scm
@@ -250,9 +250,23 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ;;
 ;; Examples:
 ;;
-;;  (glgui-timings-set! frame-period-max: 2 frame-period-min: 0.01) -- start with 110/sec down to 0.5/sec
+;;  (glgui-timings-set! frame-period-max: 2 frame-period-min: 0.01) -- start with 100/sec down to 0.5/sec
 ;;  (glgui-timings-set! frame-period-max: 0.5 frame-period-min: 0.05) -- 20/sec down to 2/sec
 ;;  (glgui-timings-set! frame-period-custom: (lambda (x) 0.01))  -- constant 100/sec
+;;
+;;  When concerned about a precise frame rate take a hint from gambit
+;;  manual wrt. `thread-sleep!` and wait until a point in time no
+;;  matter how long it took to process events and render the GUI:
+;;
+;; (glgui-timings-set!
+;;  frame-period-custom:
+;;  (let ((last-frame-time (current-time-seconds))
+;;        (frame-period 0.05))
+;;    (lambda (x) ;; ignoring `x`
+;;      (let* ((now (current-time-seconds))
+;;             (next (+ last-frame-time frame-period)))
+;;        (set! last-frame-time now)
+;;        (seconds->time next)))))
 
 (define glgui-timings-set!)
 
--- a/modules/ln_glgui/glgui.scm
+++ b/modules/ln_glgui/glgui.scm
@@ -35,7 +35,7 @@ CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
 OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 |#
-;; widget based GUI 
+;; widget based GUI
 
 ;; ----
 ;; event propagation
@@ -44,7 +44,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 (define glgui:propagate #f)
 
-(define (glgui-propagate-set! p) 
+(define (glgui-propagate-set! p)
   (set! glgui:propagate p))
 
 ;; ----
@@ -96,7 +96,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 (define (glgui-orientation-set! o)
 ;;  (let ((def (if (fx> (glgui-width-get) (glgui-height-get)) GUI_LANDSCAPE GUI_PORTRAIT)))
   (let ((def (if (> app:width app:height) GUI_LANDSCAPE GUI_PORTRAIT)))
-    (cond 
+    (cond
       ((fx= def o) (set! glgui:rotate 0))
       ((fx= o GUI_LANDSCAPE) (set! glgui:rotate 1))
       ((fx= o GUI_PORTRAIT) (set! glgui:rotate 1))
@@ -107,7 +107,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     )))
 
 (define (glgui-width-get)
-  (cond 
+  (cond
     ((fx= glgui:rotate 0) app:width)
     ((fx= glgui:rotate 1) app:height)
     ((fx= glgui:rotate 2) app:height)
@@ -115,7 +115,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
   ))
 
 (define (glgui-height-get)
-  (cond 
+  (cond
     ((fx= glgui:rotate 0) app:height)
     ((fx= glgui:rotate 1) app:width)
     ((fx= glgui:rotate 2) app:width)
@@ -142,7 +142,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
                (yofs (glgui-get g 'yofs)))
           (glPushMatrix)
           (glTranslatef (flo xofs) (flo yofs) 0.)
-          (for-each (lambda (wgt) 
+          (for-each (lambda (wgt)
             (let* ((h (glgui-widget-get g wgt 'hidden))
                    (m (glgui-widget-get g wgt 'modal))
                    (p (glgui-widget-get g wgt 'draw-handle))
@@ -157,9 +157,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
                        (glCoreClipPush 0 0 w h)
                        (glgui:renderloop wgt)
                        (glCoreClipPop)
-               )))))) 
+               ))))))
              (glgui-get g 'widget-list))
-          (glPopMatrix))) 
+          (glPopMatrix)))
     guis)))
 
 ;; render one or more guis to a window
@@ -167,7 +167,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 (define (glgui:render g1 . gx)
   (glCoreInit)  ;; setup the OpenGL pipeline
   (glPushMatrix)
-  (cond 
+  (cond
 ;;  ------ rotation code
     ((fx= glgui:rotate 1)
       (glRotatef -90. 0. 0. 1.)
@@ -178,7 +178,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     ((fx= glgui:rotate 3)
       (glRotatef 180. 0. 0. 1.)
       (glTranslatef (flo (- app:width)) (flo (- app:height)) 0.))
-;;  ------ 
+;;  ------
   )
   (apply glgui:renderloop (append (list g1) gx))
   (glPopMatrix)
@@ -187,7 +187,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 (define (glgui:for-each-done func lst)
   (let ((len (length lst)))
     (let loop ((i 0)(done #f))
-      (if (or (fx= i len) done) done 
+      (if (or (fx= i len) done) done
         (loop (fx+ i 1) (func (list-ref lst i)))))))
 
 (define (glgui:inputloop t x0 y0 . gs)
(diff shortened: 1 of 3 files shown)


0-8-15:
PS: The parameter `glgui-frame-period-max` can be used to trade of the load reduce vs. loss of responsiveness.

PS2:  As a result of this change the sleep in `loaders/common/main.c` line 183ff. could/should probably be removed as it is no longer required and just introduces a needless delay now.

mgorges:
To make sure this is documented. The expression equivalence for `when` is:

``` 
(when <test>
  <expression1>
  <expression2>
)
```
is equivalent to
```
(if <test>
  (begin
    <expression1>
    <expression2> 
  )
)
```

mgorges:
I have to admit that I don't really understand this, so you will have to provide more information on the design decisions and their potential consequences before we can consider to merge this. I.e. why you picked `0.05` and `0.5`, and why the delay increases with consecutive `EVENT_REDRAW`s instead of being fixed.

From putting a bunch of logging in there, it seems that after each `EVENT_REDRAW` you pause for 0.05 sec, then 0.10 sec, etc., until you end up running every 0.5 seconds. If there are any events other than `EVENT_REDRAW` the next `EVENT_REDRAW` redraw delay resets to `0.05` sec. This means the frame rate is now variable, between 2 and 20 Hz, while previously it was at most 100Hz, as the [loaders/common/main.c#L184](https://github.com/part-cw/lambdanative/blob/master/loaders/common/main.c#L184) sleep is `0.010` sec.

If we do this, removing the sleep from [loaders/common/main.c#L184](https://github.com/part-cw/lambdanative/blob/master/loaders/common/main.c#L184) for all apps seems fine, as `USECONSOLE` ones already didn't have them.

modules/ln_glgui/glgui.scm
```diff
@@ -230,12 +230,43 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+(define glgui-frame-period-max (make-parameter 0.5)) ;; How long to sleep at most in redraw.
+
+;; glgui-wakeup! - a thunk, which when called will immediately unblock
+;; the thread waiting in glgui-event.  Should be called if other
+;; threads notice the event loop should proceed.
+(define glgui-wakeup!)
+
+(define glgui-event
+  (let ((consecutive-redraw-count 1)
```
mgorges, line 242: If you were to change this to 0.05, and in L254 change to `(set! consecutive-redraw-count (fl+ consecutive-redraw-count step))` and change the reset in L249 to `step`, we could eliminate the multiplication in L251. It could even be changed to `0.0` in which case a redraw after an input even would be immediate.

mgorges:
I have been thinking about this a bit more and I am getting concerned about the variable and increasing delay. There are at least two pieces, which have more-real time dependencies, and rely on comparisons against `##now` to get close to accurate timings - in the worse case this change would make the timing off by 0.4999 sec. For example:
-  [modules/scheduler/scheduler.scm#L166](https://github.com/part-cw/lambdanative/blob/master/modules/scheduler/scheduler.scm#L166) and
-  [apps/DemoSerial/main.scm#L187](https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/apps/DemoSerial/main.scm#L187). 

As such we will need a way to disable this system-wide, specifically setting `step` to `0.01` and fixing `consecutive-redraw-count` to `1` in your current implementation; or setting `consecutive-redraw-count` to `0.01` and `step` to `0.0` as per the potential change I suggested above. I would then call this function in the `serial` or `scheduler` modules to revert to the current behaviour.

0-8-15:
Correct.  Standard gambit and available in many Scheme implementations.

Am Sat, 26 Sep 2020 13:20:41 -0700

> To make sure this is documented. The expression equivalence for
> `when` is:
> 
> ``` 
> (when <test>
>   <expression1>
>   <expression2>
> )
> ```
> is equivalent to
> ```
> (if <test>
>   (begin
>     <expression1>
>     <expression2> 
>   )
> )
> ```

0-8-15:
Am Sat, 26 Sep 2020 16:58:31 -0700

> I have been thinking about this a bit more and I am getting concerned
> about the variable and increasing delay. There are at least two
> pieces, which have more-real time dependencies, and rely on
> comparisons against `##now` to get close to accurate timings - in the
> worse case this change would make the timing off by 0.4999 sec. For
> example:
> -
> [modules/scheduler/scheduler.scm#L166](https://github.com/part-cw/lambdanative/blob/master/modules/scheduler/scheduler.scm#L166)
> and
> -
> [apps/DemoSerial/main.scm#L187](https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/apps/DemoSerial/main.scm#L187). 
> 
> As such we will need a way to disable this system-wide, specifically
> setting `step` to `0.01` and fixing `consecutive-redraw-count` to `1`
> in your current implementation; or setting `consecutive-redraw-count`
> to `0.01` and `step` to `0.0` as per the potential change I suggested
> above. I would then call this function in the `serial` or `scheduler`
> modules to revert to the current behaviour.

Doing so is entirely supported.  Two ways actually

1.)  Call

    (glgui-frame-period-max 0.01)

once at the start of a program which needs minimum frame rate of 100
per second.  (And remove the thread-sleep! 0.01) instead in 
https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/apps/DemoSerial/main.scm#L197 )
Then no damage is done.


2.)  The better way is actually to use the (glgui-wakeup!) and run the
time-sensitive code in it's own thread.  This would require a little
more restructuring of the code.  However improve the situation over all.

mgorges:
1) Okay, this will require a somewhat ugly hack, as we need to be able to use both `serial` and `scheduler` modules without depending on `ln_glgui`. The only way I know how to do this is this:
```
(if (function-exists? "glgui-frame-period-max")
  (eval '(glgui-frame-period-max 0.01)))
```
2) I am **NOT** debugging concurrency and timing issues in life-critical code (we use this framework to make software that gives medications to patients), so that is not something I am willing to support.
3) Can you provide some explanation for the design choices I asked about?
4) What about my suggestion to replace the multiplication with additions?

0-8-15:
Am Sat, 26 Sep 2020 15:34:12 -0700

> I have to admit that I don't really understand this, so you will have
> to provide more information on the design decisions and their
> potential consequences before we can consider to merge this. I.e. why
> you picked `0.05` and `0.5`, and why the delay increases with
> consecutive `EVENT_REDRAW`s instead of being fixed.

The picks are a bit arbitrary/guess/experience.

I admit that one could improve over the code.  So far it just happens
to handle the idle load issue me - and others -
( https://github.com/part-cw/lambdanative/issues/202 ) are concerned
with.

- With a delay greater than 0.5 the GUI starts to feel sluggish.

- 20 frames per second feel very responsive but the load is rather high

- The delay might be better an exponential back off.  The idea is that
  during idle time we better save load, but once there is interaction a
  longer delay feel sluggish again.

For added flexibility I just prepared a better version to be pushed to
the PR soon.  That includes more comments and examples.

> From putting a bunch of logging in there, it seems that after each
> `EVENT_REDRAW` you pause for 0.05 sec, then 0.10 sec, etc., until you
> end up running every 0.5 seconds. If there are any events other than
> `EVENT_REDRAW` the next `EVENT_REDRAW` redraw delay resets to `0.05`
> sec. This means the frame rate is now variable, between 2 and 20 Hz,
> while previously it was at most 100Hz, as the
> [loaders/common/main.c#L184](https://github.com/part-cw/lambdanative/blob/master/loaders/common/main.c#L184)
> sleep is `0.010` sec.
> 
> If we do this, removing the sleep from
> [loaders/common/main.c#L184](https://github.com/part-cw/lambdanative/blob/master/loaders/common/main.c#L184)
> for all apps seems fine, as `USECONSOLE` ones already didn't have
> them.

OK, did that too.  Seems to work.

0-8-15:
>     1. Okay, this will require a somewhat ugly hack, as we need to be able to use both `serial` and `scheduler` modules without depending on `ln_glgui`. The only way I know how to do this is this:
> 
> 
> ```
> (if (function-exists? "glgui-frame-period-max")
>   (eval '(glgui-frame-period-max 0.01)))
> ```

If you could send me a snippet how it would be used, I shall look into it.

Basically, a I wrote in the update above: better run this in its own thread with higher priority.  Than there is no need in those modules to learn about the existence of the GUI and its delay at all.  Which should solve this risk:

>     1. I am **NOT** debugging concurrency and timing issues in life-critical code (we use this framework to make software that gives medications to patients), so that is not something I am willing to support.

as well.  Even better than the current situation.  For the time being this tested setting a fixed timing to be identical to the current one should do the trick.

However I'd recommend to **not** use this *ugly hack* (even though it should do the trick).  Instead head to the `main.scm` or the next-level module using `serial` or ` scheduler` of the application and set the fixed timing there.

>     2. Can you provide some explanation for the design choices I asked about?

In the source now - for being closer to potential readers than these comments here.

>     3. What about my suggestion to replace the multiplication with additions?

As I had the idea of an exponential back off or customizable delay in the back of my head already, I ignored this.

Comments:

1.  I'm unsure what happens if `fl+` was applied and someone would pass a exact number in.
2.  It would spread the actual algorithm over multiple lines.  The assumed hope/rationale behind that addition was faster than multiplication.  However that only marginally true for today's processors on integers.  With floats there is nothing to gain.
3. As this is done in Scheme, the function call overhead will dominate the operation anyways.

mgorges:
1) `fl+` segfaults immediately without indication of where this happens if you were to give it an exact number.
2) I would place the `(glgui-frame-period-max 0.01)` piece at [modules/scheduler/scheduler.scm#L200](https://github.com/part-cw/lambdanative/blob/master/modules/scheduler/scheduler.scm#L200) and at [modules/serial/serial.scm#L676](https://github.com/part-cw/lambdanative/blob/master/modules/serial/serial.scm#L676), so it gets called automatically without requiring any user interaction. We cannot **change default behaviour** in a way that requires a user to modify their `main.scm`. And as we have applications that use these modules with and without use of `ln_glgui` it **must** be done in a such a way.

0-8-15:
Hm.  OK.  I have no idea how expensive the call actually is in comparison to the rest of the code.

It's just a hint.  `current-time` involves a system call after all and that's always a bit expensive.  Caching the value id `##now` and ignoring the drift is certainly a good idea if the drift is not of any concern.  Using `##now` instead will certainly do the trick.

I never came into the situation that the drift was a concern.

modules/ln_glgui/glgui.scm
```diff
@@ -264,8 +303,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
                   (apply glgui:inputloop (append (list t x0 y0) gs)))))
           (if (fx= t EVENT_REDRAW)
               (wait-for-time-or-signal!)
-              (reset-wait!))))
+              (begin
+                (thread-sleep! step)
```
mgorges, line 320: If you set frame-period-custom to a custom function, this doesn't update step so you have different timing behaviours when running in `CONSOLEAPP`s or not?

mgorges:
> t's just a hint. `current-time` involves a system call after all and that's always a bit expensive. Caching the value id `##now` and ignoring the drift is certainly a good idea if the drift is not of any concern. Using `##now` instead will certainly do the trick.

`(current-time)` is the system call which is comparatively cheap, so I believe it is the conversions to seconds that is the expensive piece.

0-8-15:
> > t's just a hint. `current-time` involves a system call after all and that's always a bit expensive. Caching the value id `##now` and ignoring the drift is certainly a good idea if the drift is not of any concern. Using `##now` instead will certainly do the trick.
> 
> `(current-time)` is the system call which is comparatively cheap, so I believe it is the conversions to seconds that is the expensive piece.

Just tried it out: yes it does it gambit.

A quick lockup in the standard (SRFI-18) returned nothing wrt. required behavior on negative times.  But I can hardly think of a reasonable person to not make that equivalent to using a zero timeout, which is required to return immediately.

mgorges:
> But I can hardly think of a reasonable person to not make that equivalent to using a zero timeout, which is required to return immediately.

That is not what I meant - it will be a time in the past

mgorges:
Wrong button, sorry.

modules/ln_glgui/glgui.scm
```diff
@@ -250,9 +250,23 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-;;  (glgui-timings-set! frame-period-max: 2 frame-period-min: 0.01) -- start with 110/sec down to 0.5/sec
+;;  (glgui-timings-set! frame-period-max: 2 frame-period-min: 0.01) -- start with 100/sec down to 0.5/sec
 ;;  (glgui-timings-set! frame-period-max: 0.5 frame-period-min: 0.05) -- 20/sec down to 2/sec
 ;;  (glgui-timings-set! frame-period-custom: (lambda (x) 0.01))  -- constant 100/sec
+;;
+;;  When concerned about a precise frame rate take a hint from gambit
+;;  manual wrt. `thread-sleep!` and wait until a point in time no
+;;  matter how long it took to process events and render the GUI:
+;;
+;; (glgui-timings-set!
```
mgorges, line 261: I think we should offer a function for people to call that does provide a reset to the old (well slightly improved) timing behaviour using this piece of code. That way we don't end up with slightly different versions of it everywhere and it can be updated in one place. So a `glgui-timings-at-10msec!` with a better name?

0-8-15: How about `(glgui-timings-set!)` with no keyword set at all?  Trivial to implement (just one `(or frame-period-max frame-period-min frame-period-custom)` to select the branch modifying the setting and resetting them in case non is given?

No second global binding, nothing.  Just return to defaults.  And only required if anything did change anything before.

mgorges: No, I want to have a function that when I call it we get the precise 10msec timing (well the improved version) back, not revert all settings to the new defaults.

0-8-15: Like this:

```
(define (glgui-timings-at-10msec!)
  (glgui-timings-set! frame-period-custom: (lambda (_) 0.01))
```
That should do the trick.

This should do it as well:

```
(define (glgui-timings-at-10msec!)
  (glgui-timings-set! frame-period-custom: #f frame-period-max: 0.01  frame-period-min: 0.01)
```
Which name should it be?

- glgui-timings-at-10msec!
- glgui-timings-traidtional!
- ...

mgorges: I was referring to the new piece you made with the `(next (+ last-frame-time frame-period)` in it. As for naming, 10msec is more clear what it does, but calling it traditional or legacy also would work - no strong preference here.

0-8-15:
Good catch!

Am Sun, 27 Sep 2020 10:56:00 -0700

> 
> 
> 
> > @@ -264,8 +303,11 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
> > OF SUCH DAMAGE.  
>                    (apply glgui:inputloop (append (list t x0 y0)
> gs))))) (if (fx= t EVENT_REDRAW)
>                (wait-for-time-or-signal!)
> -              (reset-wait!))))
> +              (begin
> +                (thread-sleep! step)
> 
> If you set frame-period-custom to a custom function, this doesn't
> update step so you have different timing behaviours when running in
> `CONSOLEAPP`s or not?
>

0-8-15:
> Am Sun, 27 Sep 2020 13:30:23 -0700 schrieb Matthias Görges <notifications@github.com>:
> Thanks for adding this - this drifts though if there as long events in between, right?
> Right. The non-drifting alternative should be the version in the comment.

Well, actually **no**!

As `##now` is set before the rendering etc. takes place this drifts only by the possible jitter between the end of the last wait and the update to `##now`.

modules/ln_glgui/glgui.scm
```diff
@@ -227,15 +227,108 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                            (customized-moment consecutive-redraw-count)
+                            (min frame-period-max-value (* consecutive-redraw-count step)))))
+              (mutex-unlock! wait-mutex wait-cv moment))
+          (reset-wait!)
+          (set! consecutive-redraw-count (fx+ consecutive-redraw-count 1))))
+    (define (glgui-event guis t x0 y0)
+      (if (and glgui:active app:width app:height)
+          (let ((gs (if (list? guis) guis (list guis))))
+            (if (fx= t EVENT_REDRAW)
+                (when (mutex-lock! wait-mutex 0)
```
mgorges, line 311: `when` is not available in Gambit 4.7.9, it was introduced only in 4.9.3. See #333. I suggest we change this to an `if` as otherwise this is a blocker to merging.

modules/ln_glgui/ln_glgui.scm
```diff
@@ -36,6 +36,23 @@ OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 |#
 
+(define-macro (define-cond-expand-feature-value v)
+  (let ((v (eval v)))
+    `(define-cond-expand-feature ,v)))
+
+(define-cond-expand-feature-value
```
mgorges, line 43: This piece should go to [modules/config/config.scm#L54](https://github.com/part-cw/lambdanative/blob/master/modules/config/config.scm#L54). You already declared the `define-macro` there, so that can be removed.

0-8-15: Would not work.  define-macro is per compilation unit in gambit.  (Some Scheme's can export them, but gambit is not among those.)

I consider to provide a patch to `gambc` instead.  However that will take some time.  In the meantime this could be a work-around.

mgorges:
You will also need to revert a701c67, as I made two commits as the Github editing tool didn't show me where to put the closing bracket.

modules/ln_glgui/ln_glgui.scm
mgorges: Okay, that is certainly another comment to put into the pull request description.

0-8-15:
> You will also need to revert [a701c67](https://github.com/part-cw/lambdanative/commit/a701c677a2491915bd090d12ec726a2474e3c353), as I made two commits as the Github editing tool didn't show me where to put the closing bracket.

The `define-macro` for gambit 7.7.9 should solve that too.

As said: I'll look into providing a patch to gambc which would allow us to take the magic back out.

mgorges:
Sorry, more questions: What is the side-effect of this on **iOS**, which uses a 1/20 sec timer-based render loop at [loaders/ios/EAGLView.m#L217](https://github.com/part-cw/lambdanative/blob/master/loaders/ios/EAGLView.m#L217), and on **android**, which would call this at [loaders/android/bootstrap.java.in#L322](https://github.com/part-cw/lambdanative/blob/b6b088f2c1cb016073a7a31c3572c52a881535ff/loaders/android/bootstrap.java.in#L322) ?

0-8-15:
> Sorry, more questions: What is the side-effect of this on **iOS**, which uses a 1/20 sec timer-based render loop at [loaders/ios/EAGLView.m#L217](https://github.com/part-cw/lambdanative/blob/master/loaders/ios/EAGLView.m#L217), and on **android**, which would call this at [loaders/android/bootstrap.javaIn#L322](https://github.com/part-cw/lambdanative/blob/b6b088f2c1cb016073a7a31c3572c52a881535ff/loaders/android/bootstrap.java.in#L322) ?

For iOS I could only guess.  For Android I have at least tested the code (i.e., the version I derived the PR from) however the same guess:

This will incur an additional delay for the application.  I.e., I should add a on-these-architectures `(glgui-timings-set! frame-period-custom: (lambda (_) 0))`.

I just did not notice any additional delay on Android.  But that's like just me noticing.

0-8-15:
While speaking of Android - with the recent move to API 26 the app does no longer receive `EVENT_IDLE` when the device is in doze mode.  I.e., 10'' after switching the display off, there is no signal anymore.  Background work stops.

Any knowledge or hint how to fix this?

mgorges:
> While speaking of Android - with the recent move to API 26 the app does no longer receive `EVENT_IDLE` when the device is in doze mode. I.e., 10'' after switching the display off, there is no signal anymore. Background work stops.

Made a new issue for this, it will get buried in this pull request when we merge it.

modules/ln_glgui/glgui.scm
```diff
@@ -325,10 +325,23 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    ;; TBD: convey the msec value to signaling code.
+    ;; switch delays to zero
+    (glgui-timings-set! frame-period-custom: no-wait))
+   (else (glgui-timings-set! frame-period-custom: wait-for-msec))))
+
+(define (glgui-timings-at-10msec!) (glgui-timings-at-msec! 0.01))
+
+(cond-expand
+ ((or android iod)
+  ;; 50 Hz should be enough - TBD: not yet effective
```
mgorges, line 342: I am not quite clear about this comment - are you proposing to test this after we merge this, instead of beforehand?

0-8-15:
> +(define (glgui-timings-at-10msec!) (glgui-timings-at-msec! 0.01))
> +
> +(cond-expand
> + ((or android iod)
                   ^

Good catch.  Aged eyes.  No, don't test this after the merge.

> +  ;; 50 Hz should be enough - TBD: not yet effective
> 
> I am not quite clear about this comment - are you proposing to test
> this after we merge this, instead of beforehand?

On Android the code is tested.  I don't have access to iOS.

The *not yet effective* refers to the fact that on these platforms
everything stays as before (for now): the event trigger is in the
platform specific code and this piece has not yet gained knowledge how
to modify the timing.

I'd propose to merge this after testing on iOS too.

Actually supporting timing changes in the platform specific code can be
done later.  (As those don't share to high idle load issue, it's IMHO
lesser priority.)

0-8-15:
> > t's just a hint. `current-time` involves a system call after all and that's always a bit expensive. Caching the value id `##now` and ignoring the drift is certainly a good idea if the drift is not of any concern. Using `##now` instead will certainly do the trick.
> 
> `(current-time)` is the system call which is comparatively cheap, so I believe it is the conversions to seconds that is the expensive piece.

This raised my suspicion.  As it would be completely against my experience.  I had to try out.  Using the gambc I built for the other patch:

1e6 times calling `(current-time)` takes 156ms.  1e6 times `((let ((x (current-time))) (lambda () (time->seconds x))))` takes 28ms. -- Less than a fifth.

No surprize here, I relax.

0-8-15:
A note wrt. timings:

With this patch we should expect a different frame rate.  Just measured the redraw time of the calculator example: takes between 1.7ms to 4ms.  With the 10ms sleep between frames there should be a drift around 3ms and about 1ms volatility.

The `timings-at-10ms!` should be able to achieve something more close to exact 10ms.

mgorges:
> 1e6 times calling `(current-time)` takes 156ms. 1e6 times `((let ((x (current-time))) (lambda () (time->seconds x))))` takes 28ms. -- Less than a fifth.

Why calling two functions instead of one is quicker seems surprising to me, but anyway it might motivate us undoing the use of ##now everywhere and going back to `(current-time-seconds)`?

mgorges:
Also, I think the user experience with the new timing is rather poor, at least in MacOS which can be seen nicely in DemoCube, but that is another issue. On Linux the default looks much smoother than on MacOS?

0-8-15:
Am Wed, 30 Sep 2020 19:43:47 -0700

> > 1e6 times calling `(current-time)` takes 156ms. 1e6 times `((let
> > ((x (current-time))) (lambda () (time->seconds x))))` takes 28ms.
> > -- Less than a fifth.  
> 
> Why calling two functions instead of one is quicker seems surprising
> to me,

There are no two functions being called.  At least not 1e6 times.

Written using named global variables this amounts to:

```
(define x (current-time))

(define (the-procedure-called-1e6-times) (time->seconds x))
```

> but anyway it might motivate us undoing the use of ##now
> everywhere and going back to `(current-time-seconds)`?

Not that fast!

The ##now is - now that #330 is merged - actually really helpful in
order to achieve more exact timings.

It keeps a record of the moment when the event was handled.

Instead of outright removing it, it might be better renamed into
something like "current-frame-time" and only be set for EVENT_REDRAW.
Other events might not need it or better continue to run with ##now as
I don't see much damage which could done by it.

mgorges:
The reason for the macOS performance issue is [loaders/macosx/SimpleOpenGLView.m#L53](https://github.com/part-cw/lambdanative/blob/master/loaders/macosx/SimpleOpenGLView.m#L53) which already runs only at 1/20 frame rate by default, and we add at minimum another 0.05sec sleep ontop of it.

0-8-15:
Am Thu, 01 Oct 2020 02:56:28 -0700

> The reason for the macOS performance issue is
> [loaders/macosx/SimpleOpenGLView.m#L53](https://github.com/part-cw/lambdanative/blob/master/loaders/macosx/SimpleOpenGLView.m#L53)
> which already runs only at 1/20 frame rate by default, and we add at
> minimum another 0.05sec sleep ontop of it.

This line:

https://github.com/part-cw/lambdanative/blob/master/modules/ln_glgui/glgui.scm#L335

should actually switch the delay off as it sets the sleep/wait time
unconditionally to zero.  How comes this is not working?

mgorges:
My current complaint is about **MacOS**, not iOS or Android for which the line above makes it zero?

PR #331, opened 2020-09-27
[REDCAP] adds redcap-project-info and optimises the redcap-import-file

karliwalti:
get fields such as get fields such as project_id, project_title, creation_time, production_time, in_production, project_language, purpose, purpose_other, project_notes, custom_record_label, secondary_unique_field, is_longitudinal, has_repeating_instruments_or_events, surveys_enabled, scheduling_enabled, record_autonumbering_enabled, randomization_enabled, ddp_enabled, project_irb_number, project_grant_number, project_pi_firstname, project_pi_lastname, display_today_now_button, missing_data_codes, external_modules
;; requires to run redcap v >10.0 on host to function properly

Issue #232, opened 2019-03-13
thread-sleep! on (my) Android: takes between 15 and 33 ms _itself_


recumbentbirder:
I try to time space sampling from the accelerometer, at the moment on Android. To get a sampling rate of 50 Hz, I need a sleep of +/- 0.02 ms (depending on data fetching from the accelerometer, but that _is_ fast).

I now found, that the call to `thread-sleep!` itself takes between 15 and 33 ms, even for very small values of the parameter:

```
(time (thread-sleep! 0.000000000000001))
```

Not so, though, for `(thread-sleep! 0`, which probably just returns immediately.

Is there an alternative to `thread-sleep!` or some hint on how to get a reliable sleep for small values?

mgorges:
To get this correct in Android is probably quite tricky due to the moving between C and Java, thus I am not surprised by your observation. If you look at the same for a desktop platform, which natively runs C-compiled code, you should be able to do much smaller periods. 

1. I am wondering, if in this case your sleeping would need to be done on the Java side, akin to the piece following [loaders/android/bootstrap.java.in#L78](https://github.com/part-cw/lambdanative/blob/master/loaders/android/bootstrap.java.in#L78) and [loaders/android/bootstrap.java.in#L116](https://github.com/part-cw/lambdanative/blob/master/loaders/android/bootstrap.java.in#L116), which specified the interval?
1. Next, we do also disable the heartbeat in Scheme, as it causes crashes in some cases. The code for this is at [languages/scm.c#L34](https://github.com/part-cw/lambdanative/blob/master/languages/scm.c#L34). It might be interesting for you to look at re-enabling it for your purposes?
1. Furthermore, I am wondering if your performance goal can actually not be met in software, but you would need to use an external hardware piece, given that Android itself is not really a realtime application? 
1. Finally, this might be something better to ask on the Gambit mailing list, see the [Gambit Wiki](http://gambitscheme.org/wiki/index.php/Main_Page) and [Github issue tracker](https://github.com/gambit/gambit/issues)

recumbentbirder:
Thanks so far! Will look into it.

For now, I use this:

```
(define (wait msec)
  (do ((t (current-milliseconds)))
    ((>= (current-milliseconds) (+ t msec)))                                 
    ))
```

Which isn't a work of beauty because of CPU load. But does what it's supposed to.

mgorges:
[...]

PR #332, opened 2020-09-27
[download] adds a threadsleep for mobile devices to keep GUI responsive

karliwalti:
analog to redcap in commit #331


mgorges:
As per #232 this probably sleeps longer than this on Android, but okay in principle.

karliwalti:
Yes, likely 0.1 s. But in case future implementations will have lower duty cycles this will still work

PR #336, opened 2020-09-28
GAMBC: backport WHEN and UNLESS

0-8-15:


libraries/libgambc/0001-Backport-macros-WHEN-and-UNLESS.patch
```diff
@@ -0,0 +1,45 @@
...
+Date: Mon, 28 Sep 2020 12:57:22 +0200
+Subject: [PATCH] Backport macros WHEN and UNLESS
+
+---
+ lib/_nonstd.scm | 22 ++++++++++++++++++++++
+ 1 file changed, 22 insertions(+)
+
+diff --git a/lib/_nonstd.scm b/lib/_nonstd.scm
+index f1db855..e66aba4 100644
+--- a/lib/_nonstd.scm
```
mgorges, line 12: The way we include patches everything above this line (or the one above it) needs to go. See [libraries/libgambc/os_shell.c.patch](https://github.com/part-cw/lambdanative/blob/master/libraries/libgambc/os_shell.c.patch) for an example.

libraries/libgambc/0002-Precompiled-C-files-for-backport-WHEN-and-UNLESS.patch
```diff
@@ -0,0 +1,10632 @@
+From f6ece71a0ada833c2cd69baf7b1d17eeea2e212a Mon Sep 17 00:00:00 2001
```
mgorges, line 1: Same problem as the other patch. Also I don't know how to sanely review a 10000 line patch as this one is not just changing a few linker line numbers. Can we reproduce this by doing the other patch and then bootstrapping it?

0-8-15: Neither do I know how to review a huge amount of generated code.

The generated `C` files are a result of these steps taken to produce
both, the source backport and the compiled files:

1. unpack gambc archive
2. `cd gambc-v4_7_9; git init; git add .; git commit -m 'initial import'`
3. `./configure; make`
4. edit `lib/_nonstd.scm` and backport the code from gambit 0.9.2
5. commit source
6. use `git format-patch` to create the first patch
7. `make from-scratch`
   This step compiles the
8. write a little test case in `testwhen.scm`:

    (display "Test when: ")
    (when (= 0 1) (display "comparsion ") (display "failed\n"))
    (unless (= 0 1) (display "comparsion ") (display "works\n"))
9. `/gsc/gsc   -:~~bin=./bin,~~lib=./lib,~~include=./include -exe testwhen.scm`
10. run `./testwhen` and verify result as `Test when: comparsion works`
11. run `./bin/gsi-script testwhen.scm`and find the same result
12. `git add` the modified `C` files
13. use `git format-patch` to create the precompiled patch
14. copy both patchs to `libgambc` directory
15. modify `make.sh`
16. check changes in and create PR

mgorges:
I redid this to compare using the actual patch from Marc, which you simplified here. That results in 1.7MB or 35000 lines of code as it also requires a more extensive update to `lib/_gambc.c` and also changes to `lib/_eval.c`, which is not great either.

0-8-15:
Am Thu, 01 Oct 2020 19:58:03 -0700

> I redid this to compare using the actual patch from Marc, which you
> simplified here.

Which "actual patch from Marc" do you refer to?

> That results in 1.7MB or 35000 lines of code as it
> also requires a more extensive update to `lib/_gambc.c` and also
> changes to `lib/_eval.c`, which is not great either.

That's the consequence of compiling the code.

Except for such things taking memory, which problem to you see, which I
am missing?

mgorges:
I was speaking of https://github.com/gambit/gambit/commit/584c8b2c40efe7111242bc72d6f84b1abaae3cd4 (you get it as a patch if you append '.patch' to the URL. Am trying your approach now and if it results in the same result will merge this tomorrow.

mgorges:
I had to do some more changes to make this patch work, so I know for certain that you didn't test this before you made the pull request. To me this would be the minimum requirement for any pull requests: ensure that at least it works on your system?

0-8-15:
Am Fri, 02 Oct 2020 02:43:35 -0700

> I had to do some more changes to make this patch work, so I know for
> certain that you didn't test this before you made the pull request.
> To me this would be the minimum requirement for any pull requests:
> ensure that at least it works on your system?


Those 16 steps clearly documented which test I ran.  Compiled and
interpreted.  Including the fact that I did not install gambc but did
that purely within the build directory.

Correct is that I could not test the lambdanative patching process with
a gambc build of LN as I don't have the resources required to do so.
Even that was mentioned (though in the context of possible build
process changes I currently could not apply).

Should I now feel sorry for having taken my spare time in an attempt to
provide you with as much as I can do to get this running?

mgorges:
Nowhere in this pull request does it say that these are untested patches. We don't need this feature, as the piece you added to ln_glgui is enough to make the changes you made there. Had this clearly been stated I would have ignored this pull request for the foreseeable future, instead of spending my nights cleaning up pull requests and try to keep things working.

0-8-15:
Am Fri, 02 Oct 2020 03:12:13 -0700

> Nowhere in this pull request does it say that these are untested
> patches.

This is technically correct as much as it is correct that I did not
claim the opposite.

Maybe that's the downside of using Github as the primary communication
medium and not having an old school mailing list for less formal,
casual conversation, asking questions etc.

The rest of this message is something which where better send via such a
mailing list (or even a via private channel).  Nothing to be archived
for eternity, but to some extend clearing these things might be
beneficial.

> Had this clearly been stated

I'm sorry.

From the conversation of the past days I'd had assumed that me pointing
out in the other thread that I am unable to follow upstream changes
right now and also [documenting in
detail](https://github.com/part-cw/lambdanative/pull/336#discussion_r496624922)
that which tests where passed the mere omission of more steps to
configure for gambc and test via some LN app had been clear enough .

> I would have ignored this pull request for the foreseeable future,
> instead of spending my nights cleaning up pull requests and try to
> keep things working.

Speaking of "had this been":  The patch itself was not tested, as I
wrote, I just copied over the patch files from `git format-patch`.  Had
the first response been "does not apply here", I'd have fixed that
saving you from your nightmares.

I'm sorry for that.

Best Regards

/Jörg

PR #337, opened 2020-09-28
CORE: better version for function-defined?

0-8-15:
marked the old version as deprecated - better check for the symbol than the string

### DIFF
--- a/modules/ln_core/misc.scm
+++ b/modules/ln_core/misc.scm
@@ -36,9 +36,11 @@ OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 |#
 
-(define (function-exists? str) 
-  (with-exception-catcher (lambda (e) #f)
-    (lambda () (with-input-from-string str (lambda () (eval (read)) #t)))
-  ))
+(define (global-variable-defined? obj) (and (symbol? obj) (##global-var? obj)))
+
+(define (function-exists? obj) ;; deprecated
+  (if (string? obj)
+      (global-variable-defined? (string->symbol obj))
+      (global-variable-defined? obj)))
 
 ;; eof


mgorges:
@0-8-15 this causes a deadlock with c391f51, as both `(and (symbol? obj) (##global-var? obj))` evaluate as #t when I call `(if (function-exists? "glgui-timings-at-10msec!") (eval '(glgui-timings-at-10msec!)))` - I will try to fix this, but otherwise I will revert this and not merge #330.

```
[ERROR] 2020-09-30 08:55:00: primordial: Unbound variable: glgui-timings-at-10msec!
[ERROR] 2020-09-30 08:55:00: primordial: Deadlock detected
[ERROR] 2020-09-30 08:55:00: primordial: Deadlock detected
```

mgorges:
Actually your preferred way, to use `(if (function-exists? 'glgui-timings-at-10msec!)` with the symbol also has the same behaviour. A quick way to test this is to add **serial** to **DemoConsole**, touch `apps/DemoConsole/main.scm` and `configure DemoConsole; make; make install;`

mgorges:
Even just including a `(display 'glgui-timings-at-10msec!)` makes `(##global-var? 'glgui-timings-at-10msec!)` **#t**

mgorges:
And of course the behaviour is different at runtime in a REPL or in GSI than when compiled ... where is ##global-var? documented in Gambit? It is declared at [gambit: lib/_kernel.scm#L4153](https://github.com/gambit/gambit/blob/2c0d6b5499e4e8d59b2be2f3c667b02d1abf2786/lib/_kernel.scm#L4153) but I can't figure this out.

peterlew:
I think I found a way to fix this by not explicitly referencing the function in question except as a string. I added this function to misc.scm:
```
(define (eval-if-exists obj)
    (if (function-exists? obj) ((eval obj))))
```
and put `(eval-if-exists "glgui-timings-at-10msec!")` in serial.scm. It doesn't crash, and `eval-if-exists` seems to work as expected when the function does exist

peterlew:
If we go with this approach, `eval-if-exists` should really only be used on strings. Calling it on a symbol will crash. Let me make some changes and propose a commit

peterlew:
Fixed in a133ab5. It is not safe to call function-exists? on a symbol, because the call itself leads to `##global-var?` returning true.

mgorges:
The strange thing with the original patch, and even the new one is this works, as it correctly evaluates to **#f**:
```
(define casefile "")
(if (function-exists? "timestamp-gettimestamp")
  ((eval 'timestamp-gettimestamp) casefile))
```
while this causes the deadlock as it evaluates as true:
```
(if (function-exists? "timestamp-gettimestamp")
  ((eval 'timestamp-gettimestamp)))
```
--> They only differ in the extra argument being in there?

0-8-15:
Am Wed, 30 Sep 2020 12:18:09 -0700
schrieb Peter Lewis <notifications@github.com>:

> Fixed in a133ab5. It is not safe to call function-exists? on a
> symbol, because the call itself leads to `##global-var?` returning
> true.

Did I already say "doesn't here"?  Or did I only want to?

0-8-15:
Am Wed, 30 Sep 2020 10:35:24 -0700

> Even just including a `(display 'glgui-timings-at-10msec!)` makes
> `(##global-var? 'glgui-timings-at-10msec!)` **#t**

Are you *sure* that you included `(display 'glgui-timings-at-10msec!)`
not `(display glgui-timings-at-10msec!)` the latter version should
actually create the global variable as a `#!unbound` value.

You you where therefore getting a '#t' and then call the supposedly
procedure in `(not safe)` mode this would explain the havoc you
encountered.  If debug mode compiled without this switch (as it it
active here) it would err out with a message.

peterlew:
I am 100% sure, as we tried many many variations trying to find out what was going on. You can try it yourself by adding `(display 'hello)` to apps/DemoConsole/main.scm, then make DemoConsole and you will find that `(##global-var? 'hello)` returns #t

0-8-15:
Am Wed, 30 Sep 2020 12:18:09 -0700
schrieb Peter Lewis <notifications@github.com>:

> Fixed in a133ab5. It is not safe to call function-exists? on a
> symbol, because the call itself leads to `##global-var?` returning
> true.

That's strange.  I can't reproduce this here on either 4.7.9 or 4.9.2 .

peterlew:
> You can try it yourself by adding `(display 'hello)` to apps/DemoConsole/main.scm, then make DemoConsole and you will find that `(##global-var? 'hello)` returns #t

This doesn't happen for you? I am using either 4.7.9 or 4.9.2 as well

0-8-15:
Am Thu, 01 Oct 2020 10:02:49 -0700
schrieb Peter Lewis <notifications@github.com>:

> > You can try it yourself by adding `(display 'hello)` to
> > apps/DemoConsole/main.scm, then make DemoConsole and you will find
> > that `(##global-var? 'hello)` returns #t  
> 
> This doesn't happen for you? I am using either 4.7.9 or 4.9.2 as well

I guess this is overcome by the patch I submitted.

The issue is a difference between compiled and interpreted code - which
I tested to work here before I submitted the patch:

In interpreted mode the global binding is not existent if nowhere
defined.  No matter how often it is referenced.

In compiled mode for any reference to `hello` anywhere gambit will
create the global var in the hope it is going to be bound to any value
*before* it it referenced.

That's the difference the latest patch catches is supposed to catch.

PR #341, opened 2020-09-30
GLGUI: fix potentialy exponential complexity - avoid use of length

0-8-15:
Note: length is O(n) - pull it out of loops where possible especially
in often invoked code like "primitives".

NB: Same warning applies to list-ref .

This patch only goes for the low hanging fruit where things should be obvious.

### DIFF
--- a/modules/ln_glgui/primitives.scm
+++ b/modules/ln_glgui/primitives.scm
@@ -61,11 +61,13 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     (glCoreEnd)))
 
 (define (glgui:draw-linestrip data color . width)
-  (if (and (list? data) (fx> (length data) 1) (fx< (length data) 200)) (begin
-    (let loop ((pts data))
-      (if (< (length pts) 2) #t (begin
-        (apply glgui:draw-line (append (car pts) (cadr pts) (list color) width))
-        (loop (cdr pts))))))))
+  (if (and (pair? data) (fx< (length data) 200)) ;; FIXME: pass the length in as parameter
+      (let loop ((pts data))
+        (or (null? pts) ;; length 0
+            (null? (cdr pts)) ;; length 1
+            (begin
+              (apply glgui:draw-line (append (car pts) (cadr pts) (list color) width))
+              (loop (cdr pts)))))))
 
 ;; ----------
 ;; box
@@ -185,6 +187,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 ;; support both legacy latex and new truetype rendering
 ;; eventually this should be stripped out for better performance
 (define (glgui:glyph-offsetx g)
+  ;; FIXME: avoid length in tight loops and frequently run code for
+  ;; being O(n), if at all possible.
   (if (fx= (length g) 5) (list-ref g 2) 0))
 (define (glgui:glyph-offsety g)
   (if (fx= (length g) 5) (list-ref g 4)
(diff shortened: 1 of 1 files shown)


mgorges:
Thanks for this - will do a bit more reviewing and then merge this. We already had a previous attempt trying to replace all `loops` with `for-each` where possible, but I don't think there are more easy and obvious places to do that; yet if we missed any that is another good things to look out for.

modules/ln_glgui/primitives.scm
```diff
@@ -61,11 +61,13 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     (glCoreEnd)))
 
 (define (glgui:draw-linestrip data color . width)
-  (if (and (list? data) (fx> (length data) 1) (fx< (length data) 200)) (begin
-    (let loop ((pts data))
-      (if (< (length pts) 2) #t (begin
-        (apply glgui:draw-line (append (car pts) (cadr pts) (list color) width))
-        (loop (cdr pts))))))))
+  (if (and (pair? data) (fx< (length data) 200)) ;; FIXME: pass the length in as parameter
```
mgorges, line 64: This future todo may need to be done in a backwards compatible way in case people have private widgets that use rely on the current argument order and this being done internally?

mgorges:
> NB: Same warning applies to list-ref .

The effect isn't all that large? If I replace a 
```
(let ((x (list-ref values 0)) (y (list-ref values 1)) (dw (list-ref values 2)))
  (p x y dw)
)
```
construct, which we use a lot, with a 
```
(let ((x (car values)) (y (cadr values)) (dw (caddr values)))
  (p x y dw)
)
```
and compile the code, you gain somewhere around 2e-5 % of performance per step?

0-8-15:
Am Wed, 30 Sep 2020 08:14:28 -0700

> > NB: Same warning applies to list-ref .  
> 
> The effect isn't all that large? If I replace a 
> ```
> (let ((x (list-ref values 0)) (y (list-ref values 1)) (dw (list-ref
> values 2))) (p x y dw)
> )
> ```
> construct, which we use a lot, with a 
> ```
> (let ((x (car values)) (y (cadr values)) (dw (caddr values)))
>   (p x y dw)
> )
> ```
> and compile the code, you gain somewhere around 2e-5 % of performance
> per step?

It's a question of complexity.  With short lists - as we have them here
- the effect is small.

But I see many places where the pattern is repeated.  Many I'd not sure
how long the list actually is.

Out of principle I'm not looking too much at the likely length of the
list.  Code is often used with cut&paste and adapted to the new use
case.  This way these issue are inherited all over the place.

Better fight the root cause.  It also make it more likely to grow
newcomers to produce better code.

0-8-15:
I wonder about backward compatibility issues at that point.

When back-porting WHEN and UNLESS I noticed a subtle difference between gambc and gambit.

Over all gambit ought to be an improvement over gambc.  *Iff* - that is - we knew no collateral damage to expect.

Is there any issue actually known, which blocks a move from gambc to gambit?

Before diving deeper into profiling/optimizing these things, I'd like to back-port/fix these things.  So far I know of only a single such subtle difference wrt. optional arguments.  If there was nothing else, I'd rather ignore gambc compatibility.

0-8-15:
Just to keep the note: profiling show that calls to `flo` account for 21 out of 36 seconds total CPU time of the calculator demo.

mgorges:
@0-8-15 Can we please limit the issues to ONE per issue or pull request? I have made #343 for your question. 

As for the use of `flo` using a lot of CPU time. - I think these are in lots of places to protect us against segfaults if someone sticks the wrong type of data into things as all `fl+` `fl*` `fl-` operations fail without location information; so almost impossible to debug for complex applications and thus **hugely expensive in terms of development time**.

0-8-15:
Am Wed, 30 Sep 2020 16:13:15 -0700

> @0-8-15 Can we please limit the issues to ONE per issue or pull
> request? I have made #343 for your question. 
> 
> As for the use of `flo` using a lot of CPU time. - I think these are
> in lots of places to protect us against segfaults if someone sticks
> the wrong type of data into things as all `fl+` `fl*` `fl-`
> operations fail without location information; so almost impossible to
> debug for complex applications and thus **hugely expensive in terms
> of development time**.


Dunno.

But I've seen that `flo` might return #f - hence I'm unsure how far
that protection goes.

When I'm coming around for performance, I'll try to macro-expand those
things away (in my experimental branch) and assess how much damage that
does.  At that point we shall see further.

PR #342, opened 2020-09-30
SCM: in debug mode: enable more debugging

0-8-15:
The gambit optimizations `(block)` and `(not safe)` are annoying in
debug mode.

Without `(block)` we gain the ability to instrument global definitions.

Without `(not safe)` we gain Scheme-level errors for segmentation faults.

In case we need something in between, we better introduce additional option (or sub-options) like `stage` - which would apply all optimizations but still include instrumentation - and/or more to remove instrumentation but keep out optimizations.

Maybe just --- hm, have a way to override these things per module ;-).  THAT would enable us to inject variants via `SETUP` or other tricks.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -52,7 +52,7 @@ compile_payload_scm()
     # -------
     # prep the compiler options
     if [ $SYS_MODE = "debug" ]; then
-      scm_opts="(declare (block)(not safe)(standard-bindings)(extended-bindings)(debug)(debug-location))"
+      scm_opts="(declare (standard-bindings)(extended-bindings)(debug)(debug-location))"
     else
       scm_opts="(declare (block)(not safe)(standard-bindings)(extended-bindings))"
     fi
--- a/modules/ln_core/ln_core.scm
+++ b/modules/ln_core/ln_core.scm
@@ -37,9 +37,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 |#
 
 (include "unit-test.scm")
+(include "flofix.scm")
 (include "color.scm")
 (include "floatstring.scm")
-(include "flofix.scm")
 (include "list.scm")
 (include "math.scm")
 (include "list-stats.scm")


mgorges:
Until we sort out the problem with #337 won't have time to think about this. Also what is the compilation time influence of this?

mgorges:
Well unless you need libraries to be compiled in debug mode, I assume you are already employing a modification to [/make.sh#L989](https://github.com/part-cw/lambdanative/blob/master/make.sh#L986), that does something like this?
```
if [ "X$SYS_MODE" = "Xdebug" ]; then
  SYS_PREFIX="${SYS_PREFIX}-debug
fi
```
I again don't quite see the merit of having module-dependent compilation flags, as unless you touch the other pieces you only suffer the additional compilation penalty once, as this approach would cache them? Yet, I am mainly wondering about the increasing diverging of `debug` and `release` versions as to me there isn't much benefit to testing anything but the production version as the compiler might screw all kinds of things up? That is unless you need to debug some very low-level feature, in which case you are far from release and the system-wide approach seems suitable?

0-8-15:
Am Wed, 30 Sep 2020 22:04:39 -0700

> Well unless you need libraries to be compiled in debug mode, I assume
> you are already employing a modification to
> [/make.sh#L989](https://github.com/part-cw/lambdanative/blob/master/make.sh#L986),
> that does something like this?
>```
> if [ "X$SYS_MODE" = "Xdebug" ]; then
>   SYS_PREFIX="${SYS_PREFIX}-debug
> fi
> ```

No, not so far.  But this would cache all things once more - this time
in debug mode, wouldn't it?

> I again don't quite see the merit of having module-dependent
> compilation flags, as unless you touch the other pieces you only
> suffer the additional compilation penalty once, as this approach
> would cache them?

Without per-module compilation flags to-debug-or-not-to-debug is a
semi-global decision.  Anything I need to touch while configured in
debug mode is recompiled with debug flags.

That's the confusing thing.  When nailing things down, I need narrow
the search.

> Yet, I am mainly wondering about the increasing
> diverging of `debug` and `release` versions as to me there isn't much
> benefit to testing anything but the production version as the
> compiler might screw all kinds of things up? That is unless you need
> to debug some very low-level feature, in which case you are far from
> release and the system-wide approach seems suitable?

For me the workflow to get to a production version is roughly:

1. Develop interactively until things are essentially working.

   During this phase I'm absolutely dependent on the ability to hook
   into global variables.  I.e., `block` mode is a killer.

2. Compile - still in debug mode and see how it works.

3. Compile in release mode and see what the compiler might have screwed
up.

0-8-15:
Am Wed, 30 Sep 2020 10:42:43 -0700

> Until we sort out the problem with #337 won't have time to think
> about this. Also what is the compilation time influence of this?

Compilation time is up during debug, sure.  No idea how much, because I
hacked the scripts to include even more debugging features gambit
includes.  In addition to those the C compiler provides.

Looking close at the effect, I believe that there is little reason to
actually enforce these options onto the source from the outside -
except to make a difference between debug mode and production without
touching the source.  They are easily included in the source file
itself (where they might be in the way during debugging).

The `(not safe)` option is actually really annoying and time consuming
in debug mode!  It turns every second issue in a segfault instead of
giving a useful hint to the reason.

mgorges:
Yes, this would make a global copy of the cache for debugging. The problem I always had when switching to debug mode is that it is possible to touch pieces and get them recompiled so you get line numbers with issues, but if you don't rebuild all modules, if your problem is deeper (it typically is for me) then I won't be told where in the end this misbehaves, as such it kind of is an all or none? 

As for having divergences between your private modifications and where the global project goes you can't really fault us for things that this may break. I can't develop for things I don't know about and I also shouldn't - I think most projects apply a if you do your own fork, you are at your own risk type of approach? 

Our willingness of considering to merge some of your changes back in really depends on what they might break or change in development behaviours we care about - #330 probably wasted 5 people-days of productivity on our side with pretty much zero benefit; I only use macOS, iOS, and in all my win32 apps have the scheduler which immediately disables most of it, and for the others this also doesn't affect android in a positive way, so no gain there either? - And this is before we will encounter future issues with the non-linear timing behaviour that I am still afraid of!

mgorges:
The removal of **(block)** is really annoying - it breaks all kinds of apps with a segfault related to (system-platform)

mgorges:
This was resolved in 7876f8e, in which I fixed the order of config going first, but there might be other related things, where we now have a module order dependency problem.

mgorges:
@0-8-15 unless you have an idea how to resolve the critical module order dependencies problem, which causes deadlocks with libgambit, or segfault with libgambc, we will reintroduce (block) to debug as we currently broke debug for everyone.

0-8-15:
Am Wed, 07 Oct 2020 22:42:07 -0700

> @0-8-15 unless you have an idea how to resolve the critical module
> order dependencies problem

Ah, at least it does what `debug` mode is supposed to do: it reveals
bugs where code depends on compiler optimizations.  ;-)

I take a look now.

>, which causes deadlocks with libgambit, or
> segfault with libgambc, we will reintroduce (block) to debug as we
> currently broke debug for everyone.

Is there a small example how to reproduce such a crash?

0-8-15:
Am Wed, 07 Oct 2020 18:22:06 -0700

> The removal of **(block)** is really annoying - it breaks all kinds
> of apps with a segfault related to (system-platform)

Current best guess: `config` needs to be the first to link, sure.

Looks like the `scm.sh` get the order wrong.  Funny this never having
been an issue when using `(block)`.  Not sure that `(block)` does
always fix that.

0-8-15:
Am Wed, 07 Oct 2020 22:42:07 -0700

> @0-8-15 unless you have an idea how to resolve the critical module
> order dependencies problem

I found it.  At least the root cause.  `make.sh` when using `add_items`
may more or less randomly reorder modules when there are transitive
dependencies.

Short time workaround: mention ALL modules in the APP's MODULES files,
do NOT trust the recursive dependency resolution via MODULES files in
other modules directories.

I have a fix for the module order so far.

But I do not understand why the `scm.sh` distinguishes `scm_coresrcs`
and `scm_auxsrcs`.  Is there a better explanation than my assumption
that this might be the remains from an early workaround of the problem
that link order is usually important for gambit but sometimes does not
matter?

0-8-15:
Am Wed, 07 Oct 2020 22:42:07 -0700

> @0-8-15 unless you have an idea how to resolve the critical module
> order dependencies problem, which causes deadlocks with libgambit, or
> segfault with libgambc, we will reintroduce (block) to debug as we
> currently broke debug for everyone.

The re-ordering might be only part of the issue.

The `eventloop.scm` does **actually** depend on `(block)` and should
also disable interrupts.

PR: #356 - I'd love to learn whether or not this might fix things by
without ensuring config is always first, before eventloop etc.

mgorges:
> Current best guess: `config` needs to be the first to link, sure.

This was fixed in 7876f8e.

mgorges:
> Short time workaround: mention ALL modules in the APP's MODULES files, do NOT trust the recursive dependency resolution via MODULES files in other modules directories.

Really don't want this, we have the option to make them include their dependencies for a reason and once that was added rely on it being present so we can update modules independently from apps.

mgorges:
> But I do not understand why the `scm.sh` distinguishes `scm_coresrcs` and `scm_auxsrcs`.

This might be that exact reason, some modules call other modules's functions and unlike C, where you can use an "include" to allow for this there isn't a standardized way to do this in Scheme?

0-8-15:
Am Thu, 08 Oct 2020 11:03:52 -0700

> > Short time workaround: mention ALL modules in the APP's MODULES
> > files, do NOT trust the recursive dependency resolution via MODULES
> > files in other modules directories.  
> 
> Really don't want this, we have the option to make them include their
> dependencies for a reason and once that was added rely on it being
> present so we can update modules independently from apps.

Things might be quite a bit out of order.  Just going through.

To summarize: this short time workaround might be out of date with my
PR.

Even though I believe the fix does not fix every possible case (which
would require a `sh` hacker to do better than me), that one still fixes
things.

The ordering algorithm up to now is seriously broken.  I recall having
been hit by that, but - not knowing where to look - found the
workaround above.

0-8-15:
Am Thu, 08 Oct 2020 11:05:49 -0700

> > But I do not understand why the `scm.sh` distinguishes
> > `scm_coresrcs` and `scm_auxsrcs`.  
> 
> This might be that exact reason, some modules call other modules's
> functions and unlike C,

No.  I Figured out that those `scm_coresrcs` implement things available
while processing embedded stuff (sounds, fonts, files).  Hence they are
pulled in front while the `scm_auxsrcs` can rely on those "magic"
definitions are already bound.

> where you can use an "include" to allow for
> this there isn't a standardized way to do this in Scheme?

No.  Scheme sets limits what's allowed and required.  Toplevel bindings
are available once bound.

However it leaves some things to the implementer.  `(block)` is one
example of such an optimization in Gambit.  In principle Scheme
requires toplevel bindings to be mutable.  And the fact that references
to a global var are either `#!unbound` or `#!void` is another fact in
gambit.

The requirement to provide gambit with the correct order of modules is
just there.  Required by gambit.

In order to know that the code does not just work by accident, we will
need to figure out how to provide a correct ordering or the modules.

mgorges:
Conceptually, all modules mentioned in a module or app are dependencies and as such need to be included before they are named themselves. This requires some wonderful brain twister to get right recursively which your pull request might have already done. Will review shortly.

mgorges:
Have tried to fix it in dcd9deb which puts all the dependencies before their parent. Well at least I hope it does.

PR #346, opened 2020-10-01
global var test

0-8-15:
This test should be better.

Obviously something is strange as this works here.


peterlew:
I don't want to expose `global-variable-defined?` because it returns #t for some unbound variables, as we've discussed. I like the use of `##global-var-ref` though. What about this?

```
(define (function-exists? obj)
  (let ((sym (if (string? obj) (string->symbol obj) obj)))
    (if (not (symbol? sym)) (raise "function-exists? must be called with a string or symbol")
      (and (##global-var? sym) (procedure? (##global-var-ref sym))))))
```

0-8-15:
Am Thu, 01 Oct 2020 09:30:42 -0700
schrieb Peter Lewis <notifications@github.com>:

> I don't want to expose `global-variable-defined?` because it returns
> #t for some unbound variables

Walking the dog is always a good idea to solve problems.

PR #347, opened 2020-10-01
NEW MODULE: generic arrays

0-8-15:
Bawden-style arrays according to SRFI-179.

Looks incredibly useful for graphic, signal processing etc.

https://srfi.schemers.org/srfi-179/srfi-179.html


mgorges:
For signal processing and graphics we use SRFI-4 homogeneous vectors like `f32vector` and `u8vector` and their relatives, with additions in [modules/ln_core/u8vector.scm](https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/modules/ln_core/u8vector.scm), and which were expanded to make some frequently used operations easier in [modules/ln_core/u8data.scm](https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/modules/ln_core/u8data.scm). Matrix operations are supported in [modules/ln_core/listlist.scm](https://github.com/part-cw/lambdanative/blob/master/modules/ln_core/listlist.scm) but are not native C-code, so they don't perform well.

Before we can start to even consider this, lets start with two big caveats: 
a) What is the license for this? If it is not BSD, we will not be able to include it.
b) This will require a large amount of new functions (>90) to be documented in the Wiki, that all need examples. We don't have resources for add that.

0-8-15:
Am Thu, 01 Oct 2020 16:32:53 -0700

> For signal processing and graphics we use SRFI-4 homogeneous vectors
> like `f32vector` and `u8vector` and their relatives, with additions
> in

The relationship to SRFI-4 etc. are discussed in the standard document.

Let me call out just one advantage: with only working atop of SRFI-4
you'll likely compute a full copy of the underlying storage and always
have to handle slicing manually.  SRFI-179 enables zero-copy
transformations.

> Before we can start to even consider this, lets start with two big
> caveats: a) What is the license for this? If it is not BSD, we will
> not be able to include it.

If I'm not mistaken than this is BSD.  Scroll down to the end the
standard document:

https://srfi.schemers.org/srfi-179/srfi-179.html


> b) This will require a large amount of new functions (>90) to be
> documented in the Wiki, that all need examples. We don't have
> resources for add that.

In the case of standards - especially publicly available and well
documented standards - it would be IMHO is is OK to simply link to the
standard itself.

mgorges:
The license is [MIT License](https://opensource.org/licenses/MIT), which would work for us. However, this is not obvious anywhere here - at minimum we should have it at the top of each of the three _SCM_ file. 

For the documentation, at minimum we would need to mention the key functions and where the documentation is available in a modules part of the wiki, as is done with [Wiki: Index of Module hpdf](https://github.com/part-cw/lambdanative/wiki/Index-of-Module-hpdf). Ideally, however, we should provide information and examples for the key functions as that is what others would expect to find in the wiki and should be done for almost all of our codebase.

0-8-15:
>Ideally, however, we should provide information and examples for the key functions as that is what others would expect to find in the wiki and should be done for almost all of our codebase.

There's a wiki page [Generalized Arrays](https://github.com/part-cw/lambdanative/wiki/Nonempty-Intervals-and-Generalized-Arrays) now.

Having double checked with the author, it seems to be a) permissible and b) the best thing to reproduce the standard itself.

mgorges:
A few remaining items:
a) README.org should be in the module folder, NOT in the LambdaNative root as it only applies to this module.
b) Renaming this module generic-arrays or generalized-arrays, instead of SRFI-179, is much more descriptive to the user.
c) To remain consistent with other documentation, and make this findable for others, we should rather make a new https://github.com/part-cw/lambdanative/wiki/Index-of-Module-generic-arrays page in the Wiki and list the functions in there; these can link to the specific sections (using the # markers ) in either SRFI-179 webpage (the preferred option), or your copy to the markup in this wiki.

gambiteer:
I like "generalized-arrays" as well as srfi-179.

mgorges:
Great thanks - will see if I can make time to run their entire test suite on all platform I have access to and otherwise merge over the weekend.

gambiteer:
I've tested the modified generic-arrays.scm with versions 4.7.9 and v4.9.3-1234-g6acd87df of Gambit.

If you use other versions of Gambit, too, I can test it on those.

mgorges:
As you are the expert on this, I will trust this as being sufficient - thanks a lot @gambiteer !

mgorges:
So before I merged, I tried two things using **DemoConsole**, in **debug** mode, against **gambit v4.9.2**: I included [@scheme-requests-for-implementation/srfi-179: /master/test-arrays.scm](https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm), minus its top 19 lines, in [apps/DemoConsole/main.scm#L40](https://github.com/part-cw/lambdanative/blob/master/apps/DemoConsole/main.scm#L40), or alternatively included it after [modules/generic-arrays/generic-arrays.scm#L111](https://github.com/part-cw/lambdanative/blob/ffbfe3ca8c3d6e7f3f7dc72e30386ab768bbc8ac/modules/generic-arrays/generic-arrays.scm#L111). Either case results in a segfault in [@scheme-requests-for-implementation/srfi-179: /master/test-arrays.scm#L376](https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm#L376), or #L377:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xffffffffffffffed)
    frame #0: 0x00000001004ec08a DemoConsole`___H_3758660784(___ps=0x000000010119a710) at 3758660784.c:424422:4
   424419	   ___SET_R2(___STK(-9))
   424420	   ___SET_R1(___PRM__3e__3d_)
   424421	   ___SET_R0(___LBL(577))
-> 424422	   ___JUMPGLONOTSAFE(___SET_NARGS(3),102,___G__25__25_every)
   424423	___DEF_SLBL(577,___L577_3758660784_23_)
   424424	   ___IF(___NOTFALSEP(___R1))
   424425	   ___GOTO(___L9359_3758660784_23_)
```
If I comment out that test, I then segfault at [@scheme-requests-for-implementation/srfi-179: /master/test-arrays.scm#L551](https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm#L551)
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xffffffffffffffed)
    frame #0: 0x00000001003a8111 DemoConsole`___H_2586204159(___ps=0x0000000101190680) at 2586204159.c:716224:4
   716221	   ___SET_R1(___FAL)
   716222	   ___SET_R0(___LBL(732))
   716223	   ___ADJFP(4)
-> 716224	   ___JUMPGLONOTSAFE(___SET_NARGS(7),249,___G_make_2d__25__25_array)
   716225	___DEF_SLBL(732,___L732_2586204159_23_)
   716226	   ___IF(___EQP(___STK(-15),___R1))
   716227	   ___GOTO(___L15761_2586204159_23_)
```
and with the same function problem in [@scheme-requests-for-implementation/srfi-179: /master/test-arrays.scm#L590](https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm#L590). And then again at [@scheme-requests-for-implementation/srfi-179: /master/test-arrays.scm#L683](https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm#L683) at which point I gave up...

0-8-15:
I forwarded the question to the actual author too.

Am Sat, 10 Oct 2020 01:39:52 -0700

> So before I merged, I tried two things using **DemoConsole**, in
> **debug** mode, against **gambit v4.9.2**: I included
> [@scheme-requests-for-implementation/srfi-179: /master/test-arrays.scm](https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm),
> minus its top 19 lines, in
> [apps/DemoConsole/main.scm#L40](https://github.com/part-cw/lambdanative/blob/master/apps/DemoConsole/main.scm#L40),
> or alternatively included it after
> [modules/generic-arrays/generic-arrays.scm#L111](https://github.com/part-cw/lambdanative/blob/ffbfe3ca8c3d6e7f3f7dc72e30386ab768bbc8ac/modules/generic-arrays/generic-arrays.scm#L111).

At this point it amounts to guess work for me.

What I did was having both a pure gsi based test and one which "load"s
just that file.  Both on linux.  No Apple here.

At a first glance, I'd guess somehow the `every` procedure is
overridden by an incompatible one.  But that#s a guess which could be
off by far.

mgorges:
I did remove lines 14-18 of https://github.com/scheme-requests-for-implementation/srfi-179/blob/master/test-arrays.scm#L14 , which do a namespace declaration in my test, but I am in the same place so this should not be needed? But every failure I had till I gave up was one of these:
```
(##namespace
   ("srfi/179#"
    ;; Internal SRFI 179 procedures that are either tested or called here.
    %%compose-indexers make-%%array %%every %%interval->basic-indexer %%interval-lower-bounds %%interval-upper-bounds %%move-array-elements %%permutation-invert %%vector-every %%vector-permute %%vector-permute->list ))
  )
```

0-8-15:
This is really something @gambiteer should be more knowledeabe  about than me.

Maybe that has something to do with the rename of the module?  I'd not expect that, but you never know.

0-8-15:
Ah, maybe I found the issue.  There is one fix I made to `ln_core/lists.scm` in order to run the tests here:

The test code requires standard [make-list](https://srfi.schemers.org/srfi-1/srfi-1.html#make-list) taking an *optional* `fill` argument.  Try this one:

```
(define (make-list n #!optional (fill 0))
  (if (and (exact-integer? n) (not (negative? n)))
      (let loop ((i n) (result '()))
        (if (> i 0)
            (loop (- i 1) (cons fill result))
            result))
      (error "make-list: The first argument must be a nonnegative integer: " n fill)))
```

gambiteer:
I'm sorry about the problems you're having.
I started with this version of `generic-arrays.scm`: https://github.com/part-cw/lambdanative/pull/347/commits/7e8ad83ad99f750aa6217a37ef1a417feda02655 which was modified from the original and made these further  modifications:  https://github.com/part-cw/lambdanative/pull/347/commits/c9b3e70b8851f19df2fe1b27be1c806ee208121e to get `test-arrays.scm` to work on `gambc-4.7.9.scm`.
In https://github.com/part-cw/lambdanative/pull/347/commits/c9b3e70b8851f19df2fe1b27be1c806ee208121e I sloppily added definitions of `make-list`, `exact`, and `inexact`, which are used in test-arrays.scm, but not in `generic-arrays.scm`.
It appears that you want to run the test suite inside the LambdaNative framework, correct?
I haven't been able to build LambdaNative yet, it fails on Ubuntu 20.04 trying to build libpng-1.6.34 with error compiling contrib/tools/pngcp.o.
I now have it building on Mac OS Catalina, but that will take a while.
How do you suggest I proceed?

gambiteer:
LambdaNative won't build with Xcode 12 on Mac OS Catalina because of the change `https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes?changes=__5`:

Clang now reports an error when you use a function without an explicit declaration when building C or Objective-C code for macOS (-Werror=implicit-function-declaration flag is on). This additional error detection unifies Clang’s behavior for iOS/tvOS and macOS 64-bit targets for this diagnostic. (49917738)

so the build fails with
```
 gcc  -DMACOSX -m64   -D___SINGLE_HOST -D___LIBRARY -D___PRIMAL -DSTANDALONE -c -o /Users/lucier/Library/Caches/lambdanative/macosx/build/945684265.o /Users/lucier/Library/Caches/lambdanative/macosx/build/945684265.c -I/Users/lucier/Library/Caches/lambdanative/macosx/include -I/Users/lucier/Library/Caches/lambdanative/macosx/include/freetype2 -I/Users/lucier/programs/lambdanative/apps/ttftool
/Users/lucier/Library/Caches/lambdanative/macosx/build/945684265.c:237:25: error: implicit declaration of function 'atol' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
  double size = (double)atol(fntsize)*0.75; // To match Latex size scale down
                        ^
1 error generated.
ERROR: failed on file /Users/lucier/Library/Caches/lambdanative/macosx/build/945684265.o
```
So I don't think I have a platform on which I can build and test LambdaNative.
So, how to proceed?

gambiteer:
I included `stdlib.h` in `/Users/lucier/programs/lambdanative/apps/ttftool/main.scm` and things are proceeding.

gambiteer:
For now, here's the small change I needed to get LambdaNative to build with Xcode 12:
```
[Bradleys-Mac-mini:~/programs/lambdanative] lucier% git diff
diff --git a/apps/ttftool/main.scm b/apps/ttftool/main.scm
index b342592..524116a 100644
--- a/apps/ttftool/main.scm
+++ b/apps/ttftool/main.scm
@@ -40,6 +40,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 (c-declare  #<<end-of-c-declare
 
+#include <stdlib.h>
 #include <stdio.h>
 #include <gd.h>
 ```

**Comment by gambiteer** [comment]
I think the way generic-arrays is packaged now in LambdaNative, the definitions of `%%every`, `make-%%array`, etc., are not included in the code for some reason.  I discovered this by hacking the files a bit and then doing
```
gsc -c -expansion test-arrays.scm > ! expansion.scm
```
where the first lines of test-arrays.scm are simply:
```
(begin
  ;; Uncomment this line to run test-arrays.scm in Gambit.
  (include "generic-arrays.scm"))
```
I think that's what's causing the crashes.
Joerg did the packaging of generic-arrays.scm for lambdanative.  I don't understand his code, though.

File: modules/generic-arrays/generic-arrays.scm [hunk 1]
```diff
@@ -0,0 +1,111 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+|#
+
+(define-macro (define-values names . body)
```
mgorges, line 25: This seems never used anywhere?

modules/generic-arrays/generic-arrays.scm
```diff
@@ -0,0 +1,111 @@
...
+    make-array array? array-domain array-getter array-dimension mutable-array? array-setter specialized-array-default-safe?
+    specialized-array-default-mutable? make-specialized-array specialized-array? array-storage-class array-indexer array-body array-safe?
+    array-elements-in-order? specialized-array-share array-copy array-curry array-extract array-tile array-translate array-permute
+    array-rotate array-reverse array-sample array-outer-product array-map array-for-each array-fold array-fold-right array-reduce array-any array-every
+    array->list list->array array-assign! array-swap! specialized-array-reshape array-ref array-set!
+    ))
+
+  (let ()
+    (include "generic-arrays-impl.scm")
+    (module-generic-arrays-end))
```
mgorges, line 108: this seems never defined?

mgorges:
> Ah, maybe I found the issue. There is one fix I made to `ln_core/lists.scm` in order to run the tests here:

I put updated that in 8b8e831 but it seems unrelated to the problems above. Not sure what happens when this is defined in here and globally?

mgorges:
> I haven't been able to build LambdaNative yet, it fails on Ubuntu 20.04 trying to build libpng-1.6.34 with error compiling contrib/tools/pngcp.o.

I don't have that version of Ubuntu available, only Ubuntu 16.04.6 LTS or Gentoo, but could make a VM if needed; where does it fail?

gambiteer:
I got lambdanative running on Mac OS Catalina with Xcode 12.
I plan to fork lambdanative and try to add a very simple module for generic-arrays, since I'm familiar with the code, and make sure I can run the tests using the lambdanative module.
Do you really like the name generic-arrays rather than generalized-arrays?  I prefer the latter, but I'll do what you like.

mgorges:
> I haven't been able to build LambdaNative yet, it fails on Ubuntu 20.04 trying to build libpng-1.6.34 with error compiling contrib/tools/pngcp.o.

I think I fixed this in 37e0c16, we had issues with stale headers in the past.

gambiteer:
If I want to blow away completely an old installation on Mac OS, what do I need to erase?
~/.lambdanative
~/Library/Caches/lambdanative
What else?  What on Linux?

mgorges:
> I plan to fork lambdanative and try to add a very simple module for generic-arrays, since I'm familiar with the code, and make sure I can run the tests using the lambdanative module.

I think it is something that @0-8-15 did when he wrapped it. When I reduce `generic-arrays.scm` to
``` 
(include "generic-arrays-impl.scm")
(include "test-arrays.scm")
```
and remove one function declared elsewhere, I can compile and your unit tests runs at least until `(palindrome? "abbc")`, where I get a deadlock with
`[ERROR] 2020-10-10 19:42:34: primordial:  '#<procedure #2>):  '(path: "girl.pgm" char-encoding: ISO-8859-1 eol-encoding: lf): (call-with-input-file: No such file or directory
`
If I provide the file I missed, it completes the tests. Gist is here: https://gist.github.com/mgorges/56feec727aaf7485593cf9582cc8f624. However, we probably need some of the backwards compatibility code for the older gambit version, so I don't simply want to commit that?

> Do you really like the name generic-arrays rather than generalized-arrays? I prefer the latter, but I'll do what you like.

No preference from me with respect to the name.

mgorges:
> If I want to blow away completely an old installation on Mac OS, what do I need to erase?
> ~/.lambdanative
> ~/Library/Caches/lambdanative
> What else? What on Linux?

`make clean` is enough if you want to get modules to be recompiled for the currently selected architecture. `make scrub` is the most aggressive form, which clears caches of all modules and libraries for all architectures. If you want to do it manually, on linux '~/.cache/lambdanative' and '../lambdanative-cache' or '$XDG_CACHE_HOME/lambdanative' are all possible locations to consider for removal.

0-8-15:
Am Sat, 10 Oct 2020 19:18:59 -0700
schrieb Bradley Lucier <notifications@github.com>:

> I got lambdanative running on Mac OS Catalina with Xcode 12.
> I plan to fork lambdanative and try to add a very simple module for
> generic-arrays, since I'm familiar with the code, and make sure I can
> run the tests using the lambdanative module. Do you really like the
> name generic-arrays rather than generalized-arrays?  I prefer the
> latter, but I'll do what you like.

Sorry, my bad.  For some reason I always miss-type that.  Dunno why.

0-8-15:
Am Sat, 10 Oct 2020 18:38:34 -0700

> +(define-macro (define-values names . body)
> 
> This seems never used anywhere?

That's true.

0-8-15:
Am Sat, 10 Oct 2020 18:38:51 -0700

> +  (let ()
> +    (include "generic-arrays-impl.scm")
> +    (module-generic-arrays-end))
> 
> this seems never defined?

It is.  In the absence of a real module system, the macro `this-module`
is used to control the scope of bindings (i.e., avoid pollution of and
conflicts with top level bindings.

It defines the module-NAME-end procedure, which updates the
prior-defined globals the exported values.

0-8-15:
Am Sat, 10 Oct 2020 19:18:59 -0700
schrieb Bradley Lucier <notifications@github.com>:

> Do you really like the
> name generic-arrays rather than generalized-arrays?  I prefer the
> latter, but I'll do what you like.

As I wrote, it's been my bad to miss-type the name.

I personally still prefer to call it srfi-179 since that hints to
source and documentation and sort of comes with the promise that there
is some portability.  The "generalized-arrays" name looks to me a bit
like "sort of but maybe not".

But I don't take a strong position wrt. the nam.

mgorges:
> It is. In the absence of a real module system, the macro `this-module` is used to control the scope of bindings (i.e., avoid pollution of and conflicts with top level bindings.

module-generic-arrays-end is never declared anywhere. Also if you want to limit the scope use a namespace as seen in some of our newer modules like [modules/minizip/minizip.scm](https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/modules/minizip/minizip.scm) and its partner in [modules/minizip/minizip%23.scm](https://github.com/part-cw/lambdanative/blob/4fa37ef4392cc218a86b7dee7813f87797a20320/modules/minizip/minizip%23.scm)

gambiteer:
Thanks, I'll look at `minizip`.

0-8-15:
Am Sun, 11 Oct 2020 11:19:07 -0700

> > It is. In the absence of a real module system, the macro
> > `this-module` is used to control the scope of bindings (i.e., avoid
> > pollution of and conflicts with top level bindings.  
> 
> module-generic-arrays-end is never declared anywhere.

It is.  From the macro `this-module`.  It's an old trick I'm using for
ages now.  Basically portable between all Scheme implementations with
little change to the macro definition itself.

I do make mistakes and overlook things.  But if that would not work,
nothing from that module would ever work.  I'd not submit such a thing.

> Also if you
> want to limit the scope use a namespace as seen in some of our newer
> modules like

I'm rarely Gambits namespaces.  Too tedious to use.

0-8-15:
Am Sat, 10 Oct 2020 19:52:13 -0700

> > I plan to fork lambdanative and try to add a very simple module for
> > generic-arrays, since I'm familiar with the code, and make sure I
> > can run the tests using the lambdanative module.  
> 
> I think it is something that @0-8-15 did when he wrapped it. When I
> reduce `generic-arrays.scm` to ``` 
> (include "generic-arrays-impl.scm")
> (include "test-arrays.scm")
> ```

Ah, looks like some declarations required only for the
`test-arrays.scm` ended up to be local definitions (which gambit would
remove during compilation for not being used) when I wrapped the
"generalized-arrays" (here without the miss-typing) in a huge `let`
form to avoid polluting the global name space restrict it to export
only those definition which are part of srfi-179.

0-8-15:
Am Sat, 10 Oct 2020 13:15:22 -0700
schrieb Bradley Lucier <notifications@github.com>:

> I think the way generic-arrays is packaged now in LambdaNative, the
> definitions of `%%every`, `make-%%array`, etc., are not included in
> the code for some reason.  I discovered this by hacking the files a
> bit and then doing ``` gsc -c -expansion test-arrays.scm > !
> expansion.scm ``` where the first lines of test-arrays.scm are simply:
> ```
> (begin
>   ;; Uncomment this line to run test-arrays.scm in Gambit.
>   (include "generic-arrays.scm"))
> ```
> I think that's what's causing the crashes.
> Joerg did the packaging of generic-arrays.scm for lambdanative.  I
> don't understand his code, though.

Ah, looks like some declarations required only for the
`test-arrays.scm` ended up to be local definitions (which gambit would
remove during compilation for not being used) when I wrapped the
"generalized-arrays" (here without the miss-typing) in a huge `let`
form to avoid polluting the global name space restrict it to export
only those definition which are part of srfi-179.

Note: The macro `this-module` is the magic, which does the rewrite.

gambiteer:
Another question: I've removed
```
(declare (standard-bindings)
         (extended-bindings)
         (block)
         (mostly-fixnum)
         (not safe))
```
from `generalized-arrays.scm`, thinking that maybe you add declarations at compile time depending on options (debug, optimize, etc.).
I suppose I could put back `(mostly-fixnum)`, it doesn't change the semantics or debugging of the program.
What are your policies about this?  Adding the first three declarations would make the `.o` a lot smaller.

mgorges:
> Another question: I've removed [...] from `generalized-arrays.scm`, thinking that maybe you add declarations at compile time depending on options (debug, optimize, etc.). [...] What are your policies about this? Adding the first three declarations would make the `.o` a lot smaller.

Correct, we use `(block)(not safe)(standard-bindings)(extended-bindings)` when not in debug mode as set in [languages/scm.sh#L61](https://github.com/part-cw/lambdanative/blob/master/languages/scm.sh#L61) and `(standard-bindings)(extended-bindings)(debug)(debug-location)` when debugging. If modules have specific needs, these are set at the top of the module. One example is [modules/eventloop/eventloop.scm#L40](https://github.com/part-cw/lambdanative/blob/master/modules/eventloop/eventloop.scm#L40), which needs `(block) (not interrupts-enabled)`. So if this module benefits from `(mostly-fixnum)` then it should be included at its start.

gambiteer:
The commit https://github.com/gambiteer/lambdanative/commit/fa2231e6b8b4b23c9ea6da96b552ee0d78fe4b3a adds a generalized-arrays module together with a test program that compiles and runs under gambc-4_7_9 and gambit-4_9_2 (as well as the current gambit github head).  I added a bit of code to the test program so it prints the number of failed tests at the end.
I don't know what bookkeeping is needed to add a module, and I forgot a few things (the copyright statement, maybe other things) but I'm getting tired and will leave it for now.
You can test it yourself if you like.

mgorges:
Thanks a lot @gambiteer - will look once done with grading or tomorrow morning and merge once we added the copyright piece and did a quick smoke test.

mgorges:
Superseeded by #369

PR #348, opened 2020-10-02
MACOSX: no wait in glgui

0-8-15:


mgorges:
I know how to disable it, but the question is whether this is the right approach. What we would want, in the end is a somewhat consistent user experience across platforms, that ideally gives the user direct control over the maximum frame rates in one place?

0-8-15:
Am Fri, 02 Oct 2020 01:51:45 -0700

> I know how to disable it, but the question is whether this is the
> right approach. What we would want, in the end, is a somewhat
> consistent user experience across platforms, that ideally gives the
> user direct control over the maximum frame rates in one place?

Ah, I see.

Yes, that would be ideal.

With the `timing-at-sec!` procedure we already have a point where this
could be hooked in.

For Android I would have a rough idea how this could be done.  When
looking into it, I found this effort blocked by the `EVENT_IDLE` not
delivered.

As there is no access to macosx and ios here, this would need
cooperation.

In medium terms we should be there.

Issue #339, opened 2020-09-28
ANDROID: No longer receiving `EVENT_IDLE` for API26+


mgorges:
While speaking of Android - with the recent move to API 26 the app does no longer receive `EVENT_IDLE` when the device is in doze mode.  I.e., 10'' after switching the display off, there is no signal anymore.  Background work stops.

Any knowledge or hint how to fix this?

_Originally posted by @0-8-15 in https://github.com/part-cw/lambdanative/pull/330#issuecomment-699986737_

0-8-15:
That's strange.  Looks like I've been using API 26 for the past year already.

However since the recent changes to the build process, which now support arm64, the code working before does now no longer.

Mysterious.

mgorges:
The new default target is 28, if you drop that down to something lower some of the legacy pieces might kick in [loaders/android/AndroidManifest.xml.in#L8](https://github.com/part-cw/lambdanative/blob/master/loaders/android/AndroidManifest.xml.in#L8). I know we used this trick initially when we had issues with the hybridapp insecure connection piece, which kind of worked but not for really new devices? That said, the Play Store will reject it so if you don't rely on that change this back to not set the target (which is what we had in the past), or set the target to the same value as the minSdkVersion?

mgorges:
I vaguely remember that we had similar issues with **_iOS_** at the beginning which we solved by become a required service, e.g. something that **plays music** or acts as a **VoIP app**, which could be declared in the manifest. https://developer.android.com/guide/components/services might have some ideas of adding a Foreground service? I can try to help but don't have an Android device available to me for testing, which makes this difficult.

0-8-15:
Looks like having fixed the `androidforegroundservice` to work again has dealt with the issue.  Getting the events again.

mgorges:
Are those commits included in #256 or is this a different piece you are developing?

PR #350, opened 2020-10-06
GLGUI: fix chat widget update.

0-8-15:
Until now the chat widget caches a pre-processed version of the
list of entries to display keyed on the `length` of the list.

Avoiding some exponential complexity during drawing, there was and
upper bound of the length maintained.  This broke once the list grew
beyond that.

This change slightly modifies the cache - and therefore renames it
within the widget.

The new caching tests ONLY whether or not the old list is the SAME in
terms of `eq?` as the new one.  It INTENSIONALY does NOT try to detect
updates to the components of the list.  This decision was made because
drawing happens much, much more frequent than updates.  It is trivial
on the update side to avoid side effects, but costly on the drawing
side to detect them.

### DIFF
--- a/modules/ln_glgui/chat.scm
+++ b/modules/ln_glgui/chat.scm
@@ -92,8 +92,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
          (y (glgui-widget-get-dyn g wgt 'y))
          (w (glgui-widget-get-dyn g wgt 'w))
          (h (glgui-widget-get-dyn g wgt 'h))
-         (strlst (glgui-widget-get-dyn g wgt 'strlst))
-         (lst (glgui-widget-get-dyn g wgt 'list))
+         (strlst-cache (glgui-widget-get-dyn g wgt 'strlst-cache))
+         (strlst (or (and (vector? strlst-cache) (vector-ref strlst-cache 1)) '()))
+         (lst (or (glgui-widget-get-dyn g wgt 'list) '()))
          (n (length lst))
          (dh (glgui-widget-get g wgt 'dh))
          (fnt (glgui-widget-get g wgt 'font))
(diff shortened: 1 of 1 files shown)


mgorges:
If one makes the reasonable assumption that for `glgui-chat` new items are typically appended at the end/bottom, instead of changing items within it, the lack of test of internal changes is probably okay. However, this wouldn't generalize for the regular `glgui-list`, which this is a special case of. Thus, I am wondering if we should provide a `glgui-chat-[updatelist]` type function that ensures both update cases: append and replacement are supported? Also, it might be useful to have a one-item comment in the constructor that points out that unlike `glgui-list`, this one cannot be simply modified by updating the `'list` list, if the length doesn't also change?

0-8-15:
Am Tue, 06 Oct 2020 15:29:37 -0700

> If one makes the reasonable assumption that for `glgui-chat` new
> items are typically appended at the end/bottom, instead of changing
> items within it, the lack of test of internal changes is probably
> okay. However, this wouldn't generalize for the regular `glgui-list`,
> which this is a special case of. Thus, I am wondering if we should
> provide a `glgui-chat-[updatelist]` type function that ensures both
> update cases: append and replacement are supported? Also, it might be
> useful to have a one-item comment in the constructor that points out
> that unlike `glgui-list`, this one cannot be simply modified by
> updating the `'list` list, if the length doesn't also change?

IMHO the assumption that the length has to change is the bad idea.

I'm re-using the chat widget to switch from one channel to another.  If
there are the same amount of messages in both chats I'm lost already.
Same goes for "keep it usable by allowing the option max a maximum
length".

The patch I made does allow both, append and update by imply updating
the 'list property.

I'd strongly recommend to NOT support a destructive update of the
list.  Reasons: a) performance, checking this case is expensive
while declaring the list immutable is simple b) error prone to implement
correctly c) I don't see it used in a destructive update case anywhere
and d) doing so would be bad practice and trivially fixed.

mgorges:
I guess I didn't quite understand the final paragraph of your initial submission correctly? It read like a warning that the new approach does NOT detect changes to internal elements within the list? Will try this out tomorrow and then merge if it handles all updates as expected by simply setting the `'list` to a new value.

0-8-15:
Am Wed, 07 Oct 2020 01:39:28 -0700

> I guess I didn't quite understand the final paragraph of your initial
> submission correctly? It read like a warning that the new approach
> does NOT detect changes to internal elements within the list? Will
> try this out tomorrow and then merge if it handles all updates as
> expected by simply setting the `'list` to a new value.

That one should be working.

Destructive updates, which keep the first pair fixed and use set-cdr!
to extend the list are not detected.

0-8-15:
Am Wed, 07 Oct 2020 01:39:28 -0700

> I guess I didn't quite understand the final paragraph of your initial
> submission correctly? It read like a warning that the new approach
> does NOT detect changes to internal elements within the list? Will
> try this out tomorrow and then merge if it handles all updates as
> expected by simply setting the `'list` to a new value.

Looks like having fixed the androidforegroundservice to work again has
dealt with the issue.

mgorges:
> Looks like having fixed the androidforegroundservice to work again has dealt with the issue.

Not quite sure what this refers to here?

0-8-15:
> > Looks like having fixed the androidforegroundservice to work again has dealt with the issue.

Wrong thead, belongs to #339.

Issue #295, opened 2020-07-23
Android mediascanner endless loop


karliwalti:
The mediascanner settings in eventloop.scm are causing android apps not to close properly. This is caused by a routine introduced in #95 
https://github.com/part-cw/lambdanative/blob/b1693b6e0bfdcb60405872963de5b205fa03aedf/modules/eventloop/eventloop.scm#L138

default should be set to true?

https://github.com/part-cw/lambdanative/blob/b1693b6e0bfdcb60405872963de5b205fa03aedf/modules/eventloop/eventloop.scm#L140

This will never become true again...

https://github.com/part-cw/lambdanative/blob/b1693b6e0bfdcb60405872963de5b205fa03aedf/modules/eventloop/eventloop.scm#L264

This will never exit. Reopening the app will lead to a grey screen. 

These lines would optionally disable the mediascanner upon termination, but it would be nicer to have a responsive mediascanner variable
```
(define (terminate . noscanner)
  (if app:android? (begin
    (let ((nm (not (if (= (length noscanner) 1) (car noscanner) #f))))
     (if nm (android-run-mediascanner))
    (android-finish)
   (if nm (let loop ()
     (if (not android-mediascanner-done?) (begin (thread-sleep! 0.1) (loop))))))))
  (if (procedure? hook:terminate) (hook:terminate))
)
```

karliwalti:
Small precision:
https://github.com/part-cw/lambdanative/blob/b1693b6e0bfdcb60405872963de5b205fa03aedf/modules/eventloop/eventloop.scm#L226

does have a callback, but that does not seem to happen.

mgorges:
I am not quite following your logic, isn't line 227 the place where android-mediascanner-done? is set to #t?

karliwalti:
Yes. It would. Does not seem to happen when mediascanner is initiated in 

https://github.com/part-cw/lambdanative/blob/b1693b6e0bfdcb60405872963de5b205fa03aedf/modules/eventloop/eventloop.scm#L261

I can't pin it down more precisely atm. Android apps hang, unless line 263f are commented.

I do get a callback when the function is called during normal operation.

peterlew:
Hi @karliwalti,

Do you have an example project involving mediascanner I can use to reproduce this issue?

Thanks,
Peter

karliwalti:
Any of the demo apps will show that behavior. 
I did build with android  21 and 22

peterlew:
[...]

PR #352, opened 2020-10-07
EVENTLOOP adds backward compatible option to disable mediascanner 

karliwalti:
Disables mediscanner on android when terminating.

Provides temporary fix for #295 for those that need it, at least for API <25 (only ones tested)


mgorges:
The variable name is confusing you call it nm, presumably for nomediascanner, yet it is true if you want it to run?

PR #356, opened 2020-10-08
EVENTLOOP: add required declarations

0-8-15:
Without `(block)` this module fails to link properly.


mgorges:
Didn't know you could do this on a per-file basis in the file, which is what you had requested originally with respect to the debug options? Thanks @0-8-15

PR #359, opened 2020-10-09
GLGUI: reduce complexity of font operations.

0-8-15:
Note: In order to keep the patch resonable sized, this patch is only a
first of a serious of changes.

So far it only proofs the concept and cracks down on the most
prominent cases profiling the app revealed.

Over time the use of `assoc` on fonts should be phased out and
replaced with properly abstracted operations to enable the underlying
implementation to be changed.

### DIFF
--- a/modules/ln_glgui/primitives.scm
+++ b/modules/ln_glgui/primitives.scm
@@ -211,6 +211,79 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 (glgui-utf8-set! #t)
 
+;;; BEGIN CODE TRANSITION
+;;;
+;;; Reducing the computational complexity of font operations improves
+;;; rendering time.
+
+(define-type ln-ttf:font
+  macros: prefix: macro-
+  desc ;; for now the legacy description of a font as a assoc-list
+  char->desc-table
+  )
+
+(define (ln-ttf:font? obj) (macro-ln-ttf:font? obj))
+
+(define find-font/desc)
+
+(define (ln-ttf:font-ref font char) ;; -> glyph
+  (cond
+   ((macro-ln-ttf:font? font) ;; TBD: leave this as the only case
+    (table-ref (macro-ln-ttf:font-char->desc-table font) char #f))
+   (else (ln-ttf:font-ref (find-font font) char))))
+
+(define (make-ln-ttf:font/desc fnt)
+  (let ((font-table (list->table
+                     (let ((double-the-key (lambda (x) (cons (car x) x))))
+                       ;; TBD: get rid of the doubling - requires changes to call sites
+                       (map double-the-key fnt)))))
+    (macro-make-ln-ttf:font fnt font-table)))
+
+(define find-font
+  ;; TBD consider amount of fonts and decide on font cache data
+  ;; structure.  Now assq-list; good for rather short lists, which
+  ;; are in line with good user interface design.
+  (let ((by-desc '() #;(make-table hash: eq?-hash))
+        (assq assq))
+    (define (%find-font/desc fnt)
+      (let ((hit (assq fnt by-desc) #;(table-ref by-desc fnt #f)))
+        (if hit (cdr hit) #; hit
+            (let ((font (make-ln-ttf:font/desc fnt)))
+              #;(table-set! by-desc fnt font)
+              (set! by-desc (cons (cons fnt font) by-desc))
+              font))))
+    (define (find-font fnt)
+      (cond
+       ((macro-ln-ttf:font? fnt) ;; new style
+        ;;
+        ;; TBD: after transition phase warn if this case is hit.
+        fnt)
+       ((and (pair? fnt) (pair? (car fnt))) ;; likely a legacy font
+        (%find-font/desc fnt))
+       (else (error "illegal font" fnt))))
+    (set! find-font/desc %find-font/desc)
+    find-font))
+
+(define (install-font-cache-backward-compatible-override!)
+  (let ((assoc.orig assoc))
+    ;; Backward compatibility: try to transparently hook into legacy
+    ;; lookups.
+    ;;
+    ;; TBD: 2020-08-23 This is an intermediate workaround to be removed
+    ;; ASAP.
+    (define (transparent-font-assoc k coll)
+      (cond
+       ((macro-ln-ttf:font? coll)
+        ;; TBD: after experimental phase warn if this case is hit.
+        (ln-ttf:font-ref coll k))
+       (else
+        ;; TBD: after transition phase error out if this case is hit.
+        (assoc.orig k coll))))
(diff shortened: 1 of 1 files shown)


0-8-15:
No comments on this one?

Note: profiling and optimization attempts are futile without this one.  It shaves the all-dominating factor from the time the rendering takes.  (For my test - the calculator - rendering takes now about *1/10th* the time it did before.

mgorges:
Sorry, it is so complex that I have not yet had the time to even look at this properly. I know that `macro-ln-ttf:font?` is never declared and again I am starting to get really risk adverse that this will break more unsuspected things. We still have a bunch of apps that don't run and finding out where the primordial exception is has been our focus for the last few days. In fact we have a bunch of Android ones that don't even load in the simulator anymore.

0-8-15:
Am Sun, 11 Oct 2020 11:16:01 -0700

> Sorry, it is so complex that I have not yet had the time to even look
> at this properly. I know that `macro-ln-ttf:font?` is never declared

It is.  See the gambit manual: `define-type` declares a type distinct
from all other types in Scheme.

It declares the `macro-ln-ttf:font?` predicate as well as accessors to
the fields in the type.

In order to not pollute the global namespace, I used the `macros:
prefix: macro-` to make them local to the compilation unit.

> We still have a bunch of apps that don't run and finding out where
> the primordial exception is has been our focus for the last few days.
> In fact we have a bunch of Android ones that don't even load in the
> simulator anymore.

That's bad.  Since most things run here, I might be helpful.

One thing I have in my local modifications for about a year, which I
never submitted, since I have no good idea (anymore) how to reproduce
the issue, is in `log:exception-handler`: I do handle
`(deadlock-exception? e)` different than the rest, just log and not go
through `log-trace`.  The comment I placed being the commented out
`lg-trace` before I forgot why reads: "Seems not to work at least for
deadlock exceptions?" .

modules/ln_glgui/primitives.scm
```diff
@@ -211,6 +211,79 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    ;; ASAP.
+    (define (transparent-font-assoc k coll)
+      (cond
+       ((macro-ln-ttf:font? coll)
+        ;; TBD: after experimental phase warn if this case is hit.
+        (ln-ttf:font-ref coll k))
+       (else
+        ;; TBD: after transition phase error out if this case is hit.
+        (assoc.orig k coll))))
+    (set! assoc transparent-font-assoc)))
```
mgorges, line 282: This overrides assoc globally, right? We use assoc in other places so what kind of errors can we expect from this?

modules/ln_glgui/primitives.scm
```diff
@@ -211,6 +211,79 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                       ;; TBD: get rid of the doubling - requires changes to call sites
+                       (map double-the-key fnt)))))
+    (macro-make-ln-ttf:font fnt font-table)))
+
+(define find-font
+  ;; TBD consider amount of fonts and decide on font cache data
+  ;; structure.  Now assq-list; good for rather short lists, which
+  ;; are in line with good user interface design.
+  (let ((by-desc '() #;(make-table hash: eq?-hash))
+        (assq assq))
```
mgorges, line 247: Are you planning to overwrite this too, or why are you copying the function here?

0-8-15:
Am Sun, 11 Oct 2020 12:53:26 -0700

> 
> 
> 
> > +(define (install-font-cache-backward-compatible-override!)  
> +  (let ((assoc.orig assoc))
> +    ;; Backward compatibility: try to transparently hook into legacy
> +    ;; lookups.
> +    ;;
> +    ;; TBD: 2020-08-23 This is an intermediate workaround to be
> removed
> +    ;; ASAP.
> +    (define (transparent-font-assoc k coll)
> +      (cond
> +       ((macro-ln-ttf:font? coll)
> +        ;; TBD: after experimental phase warn if this case is hit.
> +        (ln-ttf:font-ref coll k))
> +       (else
> +        ;; TBD: after transition phase error out if this case is hit.
> +        (assoc.orig k coll))))
> +    (set! assoc transparent-font-assoc)))
> 
> This overrides assoc globally, right? We use assoc in other places so
> what kind of errors can we expect from this?

Short answer: None.

Long answer: I understood that `assoc` is a standard binding and
widely used.  I've been careful to overwrite it in a manner where only
the case of using `assoc` with a distinct type, the one passing the
`ln-ttf:font?` predicate, will use the modified code path.

As `assoc` is a rather expensive operation anyway, the additional
dispatch (which is supposed to be an interim solution!) will do no more
harm in terms of runtime than yet another element in the list being
searched would do.

There is one exception: if one manages to pass "newstyle" font structure
to one of the five modules which declare `(extended-bindings)` as if it
still was a list, the `assoc` would complain that it expects a list.

Now given the changes I made so far, this should never happen.  The
replaced font structure is not passed anywhere outside those
procedures, which purposefully used `find-font` on the description to
retrieve the font type.

mgorges:
> One thing I have in my local modifications for about a year, which I never submitted, since I have no good idea (anymore) how to reproduce the issue, is in `log:exception-handler`: I do handle `(deadlock-exception? e)` different than the rest, just log and not go through `log-trace`. The comment I placed being the commented out `lg-trace` before I forgot why reads: "Seems not to work at least for deadlock exceptions?" .

I seems that the trace worked even for deadlock exceptions in gambit 4.7.3 but not 4.9.2? If I comment it I know it died, but no idea why, which is as difficult as the primordial deadlock loop we get?

0-8-15:
Am Sun, 11 Oct 2020 14:47:57 -0700

> > One thing I have in my local modifications for about a year, which
> > I never submitted, since I have no good idea (anymore) how to
> > reproduce the issue, is in `log:exception-handler`: I do handle
> > `(deadlock-exception? e)` different than the rest, just log and not
> > go through `log-trace`. The comment I placed being the commented
> > out `lg-trace` before I forgot why reads: "Seems not to work at
> > least for deadlock exceptions?" .  
> 
> I seems that the trace worked even for deadlock exceptions in gambit
> 4.7.3 but not 4.9.2? If I comment it I know it died, but no idea why,
> which is as difficult as the primordial deadlock loop we get?

Looks like this is actually an old issue.  I recalled it just when you
where talking about deadlock issues.  Seems that I ran into it early on.

In fact I'm aware from some discussion on the gambit mailing list, that
lambdanative makes use of gambit in a way Marc said is "not supported
by gambit".  In practice I found a way to avoid the limitation, but
this requires following a certain policy wrt. C->Scheme calls, which
the core system does not.

I'll be pondering how to fix that, but that might take some time and
will likely result in major changes.

For now I'll look into making *very* sure that no EVENT, not event
EVENT_TERMINATE is delivered to the app once it ran into system exit.

PR #360, opened 2020-10-09
LN_CORE: inline calls to `flo`

0-8-15:
Another low hanging fruit found by profiling.


mgorges:
Sorry, I don't understand this. Either this is a direct replacement, in which case you should update `fix` and `flo` directly, or it is again **some scary new** thing that we don't want. You are the only person sprinkling these `define-macro` pieces everywhere in our codebase now and the last ones when HORRIBLY wrong. As far as I know, the syntactic form `define-macro` is not part of any Scheme standard, and as such having both, and requiring the user to figure out when to use `fix` and when to use `macro-fix` is just unreasonable.

0-8-15:
> Sorry, I don't understand this.

No problem, that's what a review process is about - make coder and review both understand the code.

> Either this is a direct replacement, in which case you should update `fix` and `flo` directly

This sentence does not make much sense to me.

`fix` and `flo` are global procedures defined in `ln_core` no change about that one.

Profiling the rendering process turned out that an unreasonable large amount of time went into `color-rgba` - which in turn used `fix`a lot.

By providing a `macro-fix` which does - in effect - the same thing as `fix` does, the `gambit` compiler get's a chance to not compile calls to `fix` just to find out at runtime that all to be done is some bit testing, but instead can inline the code.

For symmetry and because it looked easy enough, I did the same thing to `flo` and then replaced all calls **within the module** with the `macro-`* version.

> or it is ... **some scary new** thing that we don't want. You are … … these `define-macro` pieces …

…

Scheme consists to the programmer of two parts, an expansion phase and a runtime phase.

`define-macro`, `define-syntax` etc. are what the pre-processor is for `C` – noting to be scared about.

Just that in Scheme this pre-processor has all the power of Scheme and works on S-expressions instead of character streams.

Not using the expansion phase cuts Scheme's power to half.  Then it's no better than JaveScript anymore.

> the last ones went HORRIBLY wrong

Sorry, I'm a human.  I make mistakes.

But - which **macro**  went horribly wrong?

> As far as I know, the syntactic form `define-macro` is not part of any Scheme standard

I'm too lazy to check those standards.  But `define-macro` used to be available in almost any Scheme implementation.  Lately it's been phased out in favor of `define-syntax`.  But `gambit`, especially the old version is not yet there.

It's not too hard to convert these few definitions from one style to the next.  Much less than the alternative to change all **your** code.

> and as such having both, and requiring the user to figure out when to use `fix` and when to use `macro-fix` is just unreasonable.

OK, this could be dealt with with a comment.  Just add one!

The `macro-fix` is only visible **within** `ln_core.scm`, but much faster.  `fix` is avail everywhere but much more costly.

`color-rgba` is the big winner wrt. the score it produces in `gprof`.

mgorges:
> For symmetry and because it looked easy enough, I did the same thing to `flo` and then replaced all calls **within the module** with the `macro-`* version.

But that doesn't hold true for `fix`? as you left those in list.scm, time.scm, and u8vector-compress.scm untouched?

0-8-15:
It holds true for `fix`.

Now obviously I've been distracted yesterday when I went over the module.  Those leftovers should be replaced too.

However looking closer, I see that these are rather irrelevant wrt. performance.  `u8vector-compress` runs a long operation in `C` the single additional call will not gain us much.

`string->seconds` is some really unexpected piece of code.  Randomly  code in the middle of a copy of `srfi-19`.  Now I found "something within srfi-19" the next bottleneck in the profile already.  That's for pointing this one out.  However I will not fix that one in the context of this PR.  It is surrounded by bad code and uses a procedure `tm:tildify` which seems not to come out of srfi-19 as it pretends to be judging from the `tm:` prefix. That one is needs to be handled to see any effect.

For `list-interpolate` I don't see what it is supposed to do.  But judging from the implementation I'd abstain from ever using it for list of more than maybe a handful of numbers.  It is worse than O(n^2).  Why should we shave small gains here?

The motivation for this pull request is purely performance related.  It reduced the remaining time my test using the calculator to display about 20 characters rendering spent in rendering by roughly **50%**.

The *remaining time* I wrote: `color-rgba` appeared in the profiler output as the next low hanging fruit only once #359 had reduced the O(fontsize*stringlenth) complexity of the string rendering.  Now it does not make much sense to name relative gains when complexity is reduced.  Those numbers are always overwhelming and vary be a huge margin depending on the factors.  In my test case the rendering takes now a bit less than a tenth of the  time it did before.

PR #361, opened 2020-10-09
CORE: flexible UTF-8 encoding error handling

0-8-15:
Until now errors in UTF-8 encoded strings are raised as an error in `string-ref '( ` *unsigned-int8* ` 1)` - which is not too helpful to locate the root cause.

With this change those errors are by default still errors, however with the string and offset of the offending byte in the error.

New is an optional second parameter `encoding-error` to **`utf8string->unicode`** - by default `#t` - which may be `#f` for *silent replacement*, `#t` for the default, which is calling the also new procedure `utf8string->unicode:on-encoding-error` with no arguments to obtain a handler to be called with two arguments, the string and offending offset and insert the value returned from the handler (if any) or a procedure to be used as if it was returned from calling `utf8string->unicode:on-encoding-error` with no arguments.

Calling `utf8string->unicode:on-encoding-error` with one argument changes the value returned by calling `utf8string->unicode:on-encoding-error` with no arguments.

As argument are valid:

- `'replace` or `#f` **:** *silent replacement*
- `'log+replace` **:** replace and *`log-error`*
- `'error`or `#t` **:** raise an error

Besides the added flexibility it is also a bit faster.

Update: also new is `(utf8string->unicode/fallback str #!key (fallback #f))` -  A version, which replaces strings causing  convertion errors with a hopefuly safe fallback of `(fallback str)`.

### DIFF
--- a/modules/ln_core/utf8string.scm
+++ b/modules/ln_core/utf8string.scm
@@ -253,7 +253,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
            (case fallback
              ((#f) (map char->integer (string->list str)))
              (else (fallback str)))
-           (handle-replloop-exception exn)))
+           (raise exn)))
      (lambda () (utf8string->unicode str (lambda (str i) (raise use-fallback)))))))
 
 (define (unicode->utf8string src)
--- a/modules/ln_core/utf8string.scm
+++ b/modules/ln_core/utf8string.scm
@@ -256,8 +256,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
            (handle-replloop-exception exn)))
      (lambda () (utf8string->unicode str (lambda (str i) (raise use-fallback)))))))
 
-(define unicode->utf8string unicode->utf8string)
-
 (define (unicode->utf8string src)
   (cond ((integer? src)
          (let ((c src))
--- a/modules/ln_core/utf8string.scm
+++ b/modules/ln_core/utf8string.scm
@@ -57,20 +57,20 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 (define utf8string? string?)
 (define utf8string=? string=?)
 (define utf8string-append string-append)
-(define utf8string-copy string-copy) 
+(define utf8string-copy string-copy)
 
 (define (utf8string-length src) (length (utf8string->unicode src)))
 
 (define (utf8string-ref s idx) (integer->utf8char (list-ref (utf8string->unicode s) idx)))
 
-(define (utf8string . cs) 
+(define (utf8string . cs)
   (let ((utf8cs (map char->utf8char cs)))
    (apply string-append utf8cs)))
 
-(define (make-utf8string n . c) 
+(define (make-utf8string n . c)
   (let ((utf8c (char->utf8char (if (fx= (length c) 1) (car c) " "))))
-    (let loop ((i 0)(s ""))
-      (if (fx= i n) s 
+    (let loop ((i 0) (s ""))
+      (if (fx= i n) s
         (loop (fx+ i 1) (string-append s utf8c))))))
 
 (define (utf8string->list s) (map integer->utf8char (utf8string->unicode s)))
@@ -108,17 +108,17 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 (define utf8string-explode (lambda (str seplst)
   (let ((utf8seplst (map char->utf8char seplst)))
     (let loop ((strlst (utf8string->list str))(tmp "")(res '()))
-      (if (= (length strlst) 0) (append res 
+      (if (= (length strlst) 0) (append res
         (if (> (utf8string-length tmp) 0) (list tmp) '()))
         (let ((chop? (member (car strlst) utf8seplst)))
           (loop (cdr strlst) (if chop? "" (utf8string-append tmp (utf8string (car strlst))))
-            (if chop? (append res (list tmp) 
+            (if chop? (append res (list tmp)
                (list (utf8string (car strlst)))) res))))))))
 
 (define (utf8string-split str sep)
   (let ((utf8sep (char->utf8char sep)))
     (let loop ((cs (utf8string->list str))(subres "")(res '()))
-      (if (= (length cs) 0) (if (> (utf8string-length subres) 0) 
+      (if (= (length cs) 0) (if (> (utf8string-length subres) 0)
         (append res (list subres)) res)
         (let* ((c (car cs))
                (split? (utf8char=? c utf8sep)))
@@ -138,20 +138,20 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
          (c1 (if (zero? pat-len) #f (utf8string-ref pattern 0)))
          (c2 (if (<= pat-len 1) #f (utf8string-ref pattern 1))))
     (cond
-     ((not c1) 0)    
+     ((not c1) 0)
      ((not c2) (utf8string-index str c1 cmp))
      (else (let outer ((pos 0))
           (cond
 	    ((> pos search-span) #f)
             ((not (cmp c1 (utf8string-ref str pos)))
-                (outer (+ 1 pos)))	
+                (outer (+ 1 pos)))
             ((not (cmp c2 (utf8string-ref str (+ 1 pos))))
                 (outer (+ 1 pos)))
             (else (let inner ((i-pat 2) (i-str (+ 2 pos)))
-               (if (>= i-pat pat-len) pos 
+               (if (>= i-pat pat-len) pos
                   (if (cmp (utf8string-ref pattern i-pat) (utf8string-ref str i-str))
                         (inner (+ 1 i-pat) (+ 1 i-str))
-                        (outer (+ 1 pos))))))))))))	
+                        (outer (+ 1 pos))))))))))))
 
 (define (utf8string-contains str pattern) (utf8string:contains str pattern utf8char=?))
 (define (utf8string-contains-ci str pattern) (utf8string:contains str pattern utf8char-ci=?))
@@ -164,7 +164,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
         (utf8newchr (char->utf8char newchr)))
     (let loop ((oldcs (utf8string->list str))(newcs '()))
       (if (= (length oldcs) 0) (list->utf8string newcs)
-        (loop (cdr oldcs) (append newcs  
+        (loop (cdr oldcs) (append newcs
           (list (if (utf8char=? (car oldcs) utf8oldchr) utf8newchr (car oldcs)))))))))
 
 (define utf8string-replace-substring string-replace-substring)
@@ -177,7 +177,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
       (let* ((moveindex (- (length first) 1))
              (newfirst (list-head first moveindex))
              (newsecond (append (list (list-ref first moveindex)) second))
-             (neww (max (utf8string-length (utf8string-mapconcat newfirst " ")) 
+             (neww (max (utf8string-length (utf8string-mapconcat newfirst " "))
                (utf8string-length (utf8string-mapconcat newsecond " ")))))
         (if (< neww bestw)
           (loop newfirst newsecond neww)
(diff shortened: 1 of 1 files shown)


modules/ln_core/utf8string.scm
```diff
@@ -177,53 +177,86 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    (with-exception-catcher
+     (lambda (exn)
+       (if (eq? exn use-fallback)
+           (case fallback
+             ((#f) (map char->integer (string->list str)))
+             (else (fallback str)))
+           (handle-replloop-exception exn)))
+     (lambda () (utf8string->unicode str (lambda (str i) (raise use-fallback)))))))
+
+(define unicode->utf8string unicode->utf8string)
```
mgorges, line 259: You define it to itself and declare it below - having just resolved order issues with the removal of (block) this doesn't quite make sense to me.

0-8-15:
Am Fri, 09 Oct 2020 13:33:21 -0700

> +(define unicode->utf8string unicode->utf8string)
> 
> You define it to itself and declare it below - having just resolved
> order issues with the removal of (block) this doesn't quite make
> sense to me.

Good catch.

I must have been gotten out of sync.  This used to define the version
without `/something` to the version I used for testing.  Eventually I
decided that `utf8string->unicode` without any trailing
'/handling-errors-or-not` should be the default.

In the code I compiled it is actually no longer there.

modules/ln_core/utf8string.scm
```diff
@@ -177,53 +177,84 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+           key
+           (case key
+             ((replace #f) utf8string->unicode:on-encoding-error-replace)
+             ((log+replace) utf8string->unicode:on-encoding-error-log+replace)
+             (else utf8string->unicode:on-encoding-error-raise-error)))))
+    (select 'error)
+    (lambda args
+      (if (null? args) handler (select (car args))))))
+
+(define (utf8string->unicode str #!optional (encoding-error #t))
```
mgorges, line 210: Can't say I fully understand this piece, but in fairness probably also not the previous version of it. However, as we don't have any unit tests on this file, how would I go about evaluating that it does as it should? @ddunsmuir uses more interesting language apps that I do, but ideally we don't want to break those only to learn something didn't go as planned?

0-8-15: > Can't say I fully understand this piece, but in fairness probably also not the previous version of it.

Do you refer t o the definition of `utf8string->unicode:on-encoding-error` or the `utf8-string->unicode`?

The former is just a setter, which takes either no arguments and returns the current handler, of takes the first arg and if it's a procedure trusts the caller and sets it as handler or is selects a pre-defined handler based on the symbol passed in.

The latter is just a slightly modified copy (the encoding error handler is new) of the code I'm using for years with utf-8.  And I do not have many things which still work when UTF-8 is broken.

> However, as we don't have any unit tests on this file, how would I go about evaluating that it does as it should?

So far I feed the old implementation with some broken UTF-8 (accidentally - I typed a "ä" into a glgui-label) which did not handle it.  The new does.  Otherwise it should rather clearly implement the URF-8 decoding.  IMHO this is a case, where it would be better to validate the source correctness than try to come up with bound to be incomplete test cases.

The error handling is another thing.  For now it just get's the job done better than the old one, because it enables users to somehow continue to work with lambdanative.  The predefined handlers seem to be compatible with the a) the old behavior, just better message b) a presumably useful case, which would log when it replaces characters and the replacing version, which at least keeps the app running.

> @ddunsmuir uses more interesting language apps that I do, but ideally we don't want to break those only to learn something didn't go as planned?

ddunsmuir:
Sorry, this is a bit beyond me. I haven't used the functions in utf8string directly before. In fact, an app I coded specifically did `(glgui-utf8-set! #f)` (in [ln_glui/primitives.scm](https://github.com/part-cw/lambdanative/blob/2e68b43106b1e1999038c423dcda7ad302dbccb8/modules/ln_glgui/primitives.scm#L203)) to go directly from utf8string to character glyphs - this was because I wrote my own code for determining characters from combined letters in Arabic characters in languages such as Sindhi. This was years ago and it's possible that changes to utf8string now would make that an unnecessary workaround. I have no current projects with test in other languages at this time - nothing to worry about breaking.

We use utf8string->unicode extensively as it is used for glgui:string->glyphs (basically all the dynamic text rendering, such as in the glgui-label). The other direction (unicode->utf8string) is also used for clipping the text when drawing. Would making sure the DemoUnicode app works be enough for testing this? It is currently working before merging this. I imagine we would find out very quickly if it broke things.

mgorges:
I tried it with [RRate](https://github.com/part-cw/LNhealth/tree/master/apps/RRate), which has a decent selection of languages and they seem fine by my side-by-side comparison. Will merge and will refer complaints.

karliwalti:
So it did break my apps with German using uiform

```
[ERROR] 2020-10-21 09:35:31: primordial: UTF-8 char encoding error "Ã\34bungen " 1
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_core/utf8string.scm line=238 col=44
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/primitives.scm line=318 col=13
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/primitives.scm line=419 col=32
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/primitives.scm line=408 col=26
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/uiform/uiform.scm line=429 col=12
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/uiform/uiform.scm line=1830 col=24
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/uiform/uiform.scm line=1838 col=10
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/glgui.scm line=151 col=36
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/glgui.scm line=145 col=11
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/glgui.scm line=183 col=3
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/ln_glgui/glgui.scm line=311 col=17
[ERROR] 2020-10-21 09:35:31: trace: /home/walterk/SoftwareRepo/lambdanative/modules/eventloop/eventloop.scm line=166 col=9
[ERROR] 2020-10-21 09:35:31: HALT
```
It can be reproduced by adding 
` (label  wrap #t text ". Es gibt drei verschiedene \303bungen welche durchgef\303<hrt werden.") `
to main.sx in WidgetDemo
Interestingly 
` (label  wrap #t text ". Es gibt drei verschiedene Übungen welche durchgeführt werden.")`
now works, which in the past would have crashed

Analog to @ddunsmuir we had to parse all special characters in the past. 

@0-8-15 Is there a way for backwards compatibility to not have to reparse all apps?

karliwalti:
> (label wrap #t text ". Es gibt drei verschiedene Übungen welche durchgeführt werden.")
now works, which in the past would have crashed

correction, that would also have worked. But writing to file not:

```
  (let ((sxstring ". Es gibt drei verschiedene Übungen welche durchgeführt werden.")
  	   (fh (if sxstring (open-output-file sxfile) #f)))
          (if  fh  (begin  (with-output-to-file sxfile (lambda () (print sxstring)))
	   (close-output-port fh))))
```

mgorges:
If I `(define sxfile "test.sx")` and change `let` to `let*` I get this to run in DemoConsole? Does it crash or where do you see the problem? The content of `cat test.sx` is ". Es gibt drei verschiedene Übungen welche durchgeführt werden." whereby the hex is 'C39C' and 'C3BC' respectively?

karliwalti:
I get '\DCbungen'

mgorges:
Are we getting a 4.9.3 vs. 4.7.2 problem again?

0-8-15:
Am Wed, 21 Oct 2020 09:25:11 -0700

> Are we getting a 4.9.3 vs. 4.7.2 problem again? 

Possible.  Unlikely, I guess.

Looks like the label content is not UTF-8.

This is definitiv not caused by the new handling - that issue was the
reason I've been looking into the code in the first place.

So, if it's a 4.9.3 vs. 4.7.2 that it has been lurking there all the
time.

0-8-15:
Am Wed, 21 Oct 2020 00:56:45 -0700
schrieb Walt K <notifications@github.com>:

> So it did break my apps with German using uiform

Interestingly.  I'm using ln_glgui - and can NOT get UTF-8

> Analog to @ddunsmuir we had to parse all special characters in the
> past. 

> @0-8-15 Is there a way for backwards compatibility to not have to
> reparse all apps?

Don't understand completely.  The old code erred out on encoding
errors, the new one too.  How comes it did not err out for you?

In order to not err out, you can use:

    (utf8string->unicode:on-encoding-error 'replace)

or

    (utf8string->unicode:on-encoding-error 'log+replace)

During initialization.  Then it should continue to work with a
replacement chat #xfffd .

You can also use

    (utf8string->unicode:on-encoding-error (lambda (str idx) 32))

To use the space (32) as replacement.

karliwalti:
Actually not. I do get the same resposes now. I was feeding the wrong string to the file. 

But back to the original issue. Can we have the uiform with following encoding working again?

`(label wrap #t text ". Es gibt drei verschiedene \303�bungen welche durchgef\303<hrt werden.")`

mgorges:
Debugging this at the uiform level is unnecessarily complicated because it only calls `utf8string->unicode` via `glgui:string->glyphs` in some very looped way. Can you provide a string you feed into `utf8string->unicode` with the previous `list` result you expect? Then we can see if this is possible? Feeding your string example into this now gives a `(46 32 69 115 32 103 105 98 116 32 100 114 101 105 32 118 101 114 115 99 104 105 101 100 101 110 101 32 239 . #!void)` which is a lot shorter than the provided string so that already fails in the old version (from October 10 in the repository).

0-8-15:
Am Thu, 22 Oct 2020 11:01:52 -0700

> Debugging this at the uiform level is unnecessarily complicated
> because it only calls `utf8string->unicode` via
> `glgui:string->glyphs` in some very looped way. Can you provide a
> string you feed into `utf8string->unicode` with the previous `list`
> result you expect? Then we can see if this is possible? Feeding your
> string example into this now gives a `(46 32 69 115 32 103 105 98 116
> 32 100 114 101 105 32 118 101 114 115 99 104 105 101 100 101 110 101
> 32 239 . #!void)`
         ^^^^^^^^
WAIT!  That looks like the culprit.  Should not happen!

karliwalti:
So this is my string
[uebungen.txt](https://github.com/part-cw/lambdanative/files/5428610/uebungen.txt)

karliwalti:
in current master:
(unicode->utf8string (utf8string->unicode "Es gibt drei verschiedene \303bungen welche durchgef\303<hrt werden."))
UTF-8 char encoding error "Es gibt drei verschiedene �bungen welche durchgef�<hrt werden." 27
> "Es gibt drei verschiedene \303\242ungen welche durchgef\303\274hrt werden."

In current master with checking out old utf8string.scm (pre-merge of this request)
(unicode->utf8string (utf8string->unicode "Es gibt drei verschiedene \303bungen welche durchgef\303<hrt werden."))
"Es gibt drei verschiedene \303\242ungen welche durchgef\303\274hrt werden."

mgorges:
> WAIT! That looks like the culprit. Should not happen!

Sorry I don't think that is right, that with the older version, not the new one?

mgorges:
I had done this on the pre-merge commit:
```
(define str ". Es gibt drei verschiedene \303�bungen welche durchgef\303<hrt werden.")
(for-each display (list (u8vector->list (string->u8vector str)) "\n"))
(for-each display (list (utf8string->unicode str) "\n"))
```
which gives
```
(46 32 69 115 32 103 105 98 116 32 100 114 101 105 32 118 101 114 115 99 104 105 101 100 101 110 101 32 195 239 191 189 98 117 110 103 101 110 32 119 101 108 99 104 101 32 100 117 114 99 104 103 101 102 195 60 104 114 116 32 119 101 114 100 101 110 46)
(46 32 69 115 32 103 105 98 116 32 100 114 101 105 32 118 101 114 115 99 104 105 101 100 101 110 101 32 239 . #!void)
```

karliwalti:
I think it seems to still do the job as desired, but throws an error. would a warning be better to continue executing the code?

PR #362, opened 2020-10-10
enable C profiling and debugging of Scheme code.

0-8-15:
This might be helpful to track the breakage under mac down.

Also likely helpful to reveal bottlenecks by profiling.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -56,17 +56,18 @@ compile_payload_scm()
     # prep the compiler options
     if [ $SYS_MODE = "debug" ]; then
       scm_opts="(declare (standard-bindings)(extended-bindings)(debug)(debug-location))"
+      payload_cdefs="$payload_cdefs -D___LIBRARY -D___PRIMAL"
     else
       scm_opts="(declare (block)(not safe)(standard-bindings)(extended-bindings))"
+      payload_cdefs="$payload_cdefs -D___SINGLE_HOST -D___LIBRARY -D___PRIMAL"
     fi
     scm_opts="${scm_opts}(define-cond-expand-feature $SYS_PLATFORM)"
+    if [ `is_standalone_app` = "yes" ]; then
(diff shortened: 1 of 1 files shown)


mgorges:
That sounds like a wonderful addition. Would you mind adding a paragraph to https://github.com/part-cw/lambdanative/wiki/Debugging about how you use `gprof` on Linux, as it doesn't run on macOS, where one has to use Google's `pprof` or another solution instead?

0-8-15:
Done.

Am Sat, 10 Oct 2020 16:19:01 -0700

> That sounds like a wonderful addition. Would you mind adding a
> paragraph to https://github.com/part-cw/lambdanative/wiki/Debugging
> about how you use `gprof` on Linux, as it doesn't run on macOS, where
> one has to use Google's `pprof` or another solution instead?

mgorges:
Thanks.

0-8-15:
Am Sat, 10 Oct 2020 16:19:01 -0700

> That sounds like a wonderful addition. Would you mind adding a
> paragraph to https://github.com/part-cw/lambdanative/wiki/Debugging

That page details how to debugging unused variables:
https://github.com/part-cw/lambdanative/wiki/Debugging#undefined-variable-example-using-debug-mode

That's been a major issue for me all the time (being slightly visually
impaired it happens to me all the time).  I tried this out, but no
effect.  Does this work for you?

mgorges:
> That's been a major issue for me all the time (being slightly visually impaired it happens to me all the time). I tried this out, but no effect. Does this work for you?

On Linux with **libgambc** it does - see Gist: https://gist.github.com/mgorges/9727d3eb702a408b559c0c8eac8295c4

mgorges:
On MacOS with **libgambit** it does NOT, and makes one of those wonderful primordial loop things - see Gist: https://gist.github.com/mgorges/320fd2bd2ea7b98cf4283028a82aeed3

PR #364, opened 2020-10-10
CORE: optimize time formatting.

0-8-15:
Removed use of '`length`, less memory hungry and computationally expensive string conversion, pull handling of optional arguments up, avoid needless string conversions.

With this change the date formatting related entry disappears from the top list in `gprof(1)`.

Use of `flo`is now again up, right after the **sum of all** from `glcore`.


modules/ln_core/time.scm
```diff
@@ -1485,19 +1485,27 @@ end-of-c-declare
...
+   ((string-contains str "%")
+    (let ((result (string-copy str)))
+      (do ((i 0 (fx+ i 1)))
+          ((fx= i (string-length result))
+           result)
+        (if (eqv? (string-ref result i) #\%)
+            (string-set! result i #\~)))))
+   (else str)))
+
+(define (string->seconds~ str fmt . tz0)
```
karliwalti, line 1502: is this tilda needed?

karliwalti:
`string->seconds` is no longer defined?

breaks my compiling

0-8-15:
Am Mon, 12 Oct 2020 09:00:44 -0700
schrieb Walt K <notifications@github.com>:

> `string->seconds` is no longer defined?
> 
> breaks my compiling  

It's still defined here.

0-8-15:
Am Mon, 12 Oct 2020 09:00:16 -0700
schrieb Walt K <notifications@github.com>:

> @karliwalti commented on this pull request.
> 
> 
> 
> > -  (let* ((date (string->date str (tm:tildify fmt)))  
> +;;
> +;, Why actually?  And why at runtime?
> +(define (ln:tildify str)
> +  (cond
> +   ((equal? str "") str)
> +   ((string-contains str "%")
> +    (let ((result (string-copy str)))
> +      (do ((i 0 (fx+ i 1)))
> +          ((fx= i (string-length result))
> +           result)
> +        (if (eqv? (string-ref result i) #\%)
> +            (string-set! result i #\~)))))
> +   (else str)))
> +
> +(define (string->seconds~ str fmt . tz0)
> 
> is this tilda needed?

I would vote to remove it.  It's been incomplete from the start anyway:
how to include a #\~ in the resulting string?

Just did not want to change anything functional.

karliwalti:
`(define (string->seconds~ str fmt . tz0)` this tilda was added.

mgorges:
> would vote to remove it. It's been incomplete from the start anyway: how to include a #\~ in the resulting string?

This is needed as we have some code that uses things like`%y%m%d`, while the updated module that @clpetersen added many years ago uses `~y~m~d` internally.

mgorges:
> `string->seconds` is no longer defined? breaks my compiling

Sorry about that I had unit-tested it and ran a bunch of apps, but I guess we don't have a test for it. Added back in 3b9c6bf

Issue #15, opened 2013-09-13
Latest commit bf53628ef10aeceea0232519e5015f2894d83e33 breaks build


fungiblecog:
Was building fine on linux. Latest pull has caused breakage:

./make.sh: 1: eval: /home/fungiblecog/lambdanative/lambdanative-cache/linux/bin/subtool: not found

PR #366, opened 2020-10-12
CORE: try to avoid calling into Scheme after exit.

0-8-15:
With libgambit this errors out for me for unbound variables like this:

  Code: `(unbound display)` Error:

  Thread "primordial": (#!unbound '#<procedure #15 display>): Operator is not a PROCEDURE

Unfortunately ___cleanup(); in libgambit will still raise a deadlock
exception in log-trace.

### DIFF
--- a/languages/scm.c
+++ b/languages/scm.c
@@ -29,6 +29,7 @@ int debug_settings = ___DEBUG_SETTINGS_INITIAL;
 
 void system_init();
 
+static int lambdanative_exit_call_count = -1;  /* System running when 0 */
 void lambdanative_payload_setup()
 {
   DMSG("lambdanative_payload_setup [scm]");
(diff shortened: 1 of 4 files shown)


mgorges:
I like the idea, however, we would break some existing feature, which is pointing to the right place for the deadlock in the older gambit version. So maybe we should only selectively disable the trace piece for the new version?

0-8-15:
Am Tue, 13 Oct 2020 20:11:49 -0700

> I like the idea, however, we would break some existing feature, which
> is pointing to the right place for the deadlock in the older gambit
> version. So maybe we should only selectively disable the trace piece
> for the new version?

I would be OK with that.

The real solution would be to update the log-trace for new gambit.  But
I don't have the resources for that.

mgorges:
> I would be OK with that. 

Using a `(cond-expand (gambit-c` type of approach or what is the best test for this?

> The real solution would be to update the log-trace for new gambit. But I don't have the resources for that.

Agree, but this is calling some gambit-internal piece and as such getting documentation for it might be tricky?

0-8-15:
Am Wed, 14 Oct 2020 07:47:04 -0700

> > I would be OK with that.   
> 
> Using a `(cond-expand (gambit-c` type of approach or what is the best
> test for this?

That's what I would do.

> > The real solution would be to update the log-trace for new gambit.
> > But I don't have the resources for that.  
> 
> Agree, but this is calling some gambit-internal piece and as such
> getting documentation for it might be tricky?

Yeah, the documentation is all there: the code.

With some luck we could get help on the gambit mail list.

PR #367, opened 2020-10-12
BUILD: export well known build parameters to gambit compiler

0-8-15:
This is the smallest change and most generic way I could come up with to tell the Scheme compiler more about the target environment.

Note that the `export`ed list might need to grow or being cut down.  Right now it's just a suggestion.  It will also require some care, as it is now "more-or-less-documented" once those build parameters where exported.

`SYS_ANDROIDAPI` looks right now like the minimum to make with #256 available on both older and newer versions.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -129,7 +129,11 @@ compile_payload_scm()
           gsc_processing=""
         fi
 	# if [ $SYS_VERBOSE ]; then gsc_processing="$gsc_processing -expansion"; fi
-        veval "$SYS_GSC -:~~tgt=${SYS_PREFIX} -prelude \"$scm_opts\" -c -o $scm_ctgt $gsc_processing $scm_hdr $scm_src"
+        (
+          # exported list of "well known" build parameters
+          export APP_ABI SYS_PREFIX SYS_PATH SYS_ROOT SYS_ANDROIDAPI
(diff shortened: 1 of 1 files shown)


mgorges:
Before we set these all in stone, as APP_ABI is only used in ANDROID, so maybe we should rename it to ANDROID_ABI or SYS_ANDROID_ABI to be consistent? Also what do we need SYS_ROOT for?

0-8-15:
Am Tue, 13 Oct 2020 19:59:31 -0700

> Before we set these all in stone, as APP_ABI is only used in ANDROID,
> so maybe we should rename it to ANDROID_ABI or SYS_ANDROID_ABI to be
> consistent?

Agreed.  Could be (for now) even being done locally at the spot.

> Also what do we need SYS_ROOT for?

There an idea floating in my head, which would make this and a
to-be-cleaned-up similar small change here the only change required to:

a) get the jscheme support working (I know you are short on time, but
if you had a quick look at it, or remember your last assessment), this
module could be:

  - all ever required in terms of JNI (maybe with the exception of very time-sensitive things)

  - enable switching between LN's glgui and using webview and other
    native Java stuff from just a few lines of Scheme

  - switch layouts in the startup, thus possibly replacing special
    cases like hybrid-app handling

b) I need to compile another `.so` for use with Android.

c) I need to find all the input components from within gambit.  It
might be nowhere else needed.  But for me it would be an empowering
feature.  Likely others would like this too, once there's an example
use case.

PR #369, opened 2020-10-12
Add module generalized-arrays

gambiteer:
This adds the module source.
I added a wiki page for documentation: https://github.com/part-cw/lambdanative/wiki/Index-of-Module-generalized-arrays
I figured that wiki pages are easy to correct.
I don't know what else is needed to incorporate the module into LambdaNative.


mgorges:
Wonderful thanks a lot - give us another day for testing and then I'll merge it.

gambiteer:
I don't yet understand LambdaNative's build process, but let me ask: are the `include`s at lines 35, 36, and 39 of `generalized-arrays.scm` not a problem too?

mgorges:
They are not - I think it has to do with the use in the macros, but why this is the case seems to be above my pay grade. The version I just merged compiles on Linux and macOS for those targets, and the unit tests passes on both platforms. I can also compile it for android; didn't test iOS or win32 specifically, but I don't expect them to behave differently in this respect.

PR #375, opened 2020-10-19
UIFORM SLIDER fix for default display and modular slider colour

karliwalti:
fix: does not set value to defaultvalue (as otherwise the sliderbox would display as selected before any user interaction)
feature: adds `selcolour` to define the coulour of the sliderbox


mgorges:
From trying this in WidgetDemo the number in the slider is pretty much invisible now, as it is white in a partially-transparent white box? That doesn't seem right?

karliwalti:
Ok, added font color of value also to configurable options. 

Also added slider to objects in the DemoUIform

mgorges:
Will review tomorrow. As an aside, https://github.com/part-cw/LNhealth/blob/master/apps/WidgetDemo/sandbox/main.sx has quite a few more examples, which is pointed out in [apps/DemoUIForm/main.scm#L80](https://github.com/part-cw/lambdanative/blob/master/apps/DemoUIForm/main.scm#L80) - maybe they need to be consolidated?

karliwalti:
sorry, I had accidently pushed too much...

PR #377, opened 2020-10-20
CAMERA (Android): request permissions for API > 22 and allow for exte…

peterlew:
…rnal storage URI

The Override function onRequestPermissionsResult is needed in bootstrap.java.in to prevent apps from hanging after asking permissions. A StrictMode.VmPolicy.Builder is needed in the camera module to get the URI of a file outside the app's storage, as per this thread: https://stackoverflow.com/questions/48117511/exposed-beyond-app-through-clipdata-item-geturi


mgorges:
Superseded by #380

PR #389, opened 2020-11-02
BUILD: enable final link via C++

0-8-15:
It also links some files into a position where they are expected.

No idea how to do this the right way(TM).

### DIFF
--- a/targets/win32/build-binary
+++ b/targets/win32/build-binary
@@ -87,7 +87,7 @@ if [ `is_standalone_app` = "yes" ]; then
 else
   if [ `is_gui_app` = yes ]; then
     veval "$SYS_CC -I$SYS_PREFIX/include \
-      $cflag_additions -mwindows win32_microgl.c main.c icon.o -o $tgt \
+      $cflag_additions -mwindows -fpermissive win32_microgl.c main.c icon.o -o $tgt \
       -L$SYS_PREFIX/lib $ldflag_additions -lpayload -lglu32 -lopengl32 -lgdi32 -lwinmm -lwsock32 -lws2_32 -lrpcrt4 -lcrypt32 -lm"
   else
     veval "$SYS_CC -I$SYS_PREFIX/include \
--- a/targets/win32/host_linux
+++ b/targets/win32/host_linux
@@ -38,6 +38,11 @@
 win_sysroot=`ls -1d ${WIN32SDK}/i?86-*mingw32* | head -n 1`
 win_cross=`ls -1 ${WIN32SDK}/bin/*mingw32*-gcc | head -n 1 | sed 's/gcc$//'`
 
+test -f ${SYS_PREFIX}/include/WinSock2.h || ln -s $win_sysroot/include/winsock2.h ${SYS_PREFIX}/include/WinSock2.h
+test -f ${SYS_PREFIX}/include/WS2tcpip.h || ln -s $win_sysroot/include/ws2tcpip.h ${SYS_PREFIX}/include/WS2tcpip.h
+test -f ${SYS_PREFIX}/include/Windows.h || ln -s $win_sysroot/include/windows.h ${SYS_PREFIX}/include/Windows.h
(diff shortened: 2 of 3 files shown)


targets/win32/host_linux
```diff
@@ -38,6 +38,11 @@
 win_sysroot=`ls -1d ${WIN32SDK}/i?86-*mingw32* | head -n 1`
 win_cross=`ls -1 ${WIN32SDK}/bin/*mingw32*-gcc | head -n 1 | sed 's/gcc$//'`
 
+test -f ${SYS_PREFIX}/include/WinSock2.h || ln -s $win_sysroot/include/winsock2.h ${SYS_PREFIX}/include/WinSock2.h
```
mgorges, line 41: Why are you doing this instead of adding another `-I$win_sysroot/include` in targets/win32/build-binary#L89? Also, does this filename case matter?

targets/win32/build-binary
```diff
@@ -87,7 +87,7 @@ if [ `is_standalone_app` = "yes" ]; then
 else
   if [ `is_gui_app` = yes ]; then
     veval "$SYS_CC -I$SYS_PREFIX/include \
-      $cflag_additions -mwindows win32_microgl.c main.c icon.o -o $tgt \
+      $cflag_additions -mwindows -fpermissive win32_microgl.c main.c icon.o -o $tgt \
```
mgorges, line 90: Shouldn't the `-fpermissive` also only be applied when using a c++ compiler, as I thought this was only needed if you use non-conforming code?

0-8-15:
Am Wed, 04 Nov 2020 00:44:39 -0800

> 
> 
> 
> > @@ -38,6 +38,11 @@  
>  win_sysroot=`ls -1d ${WIN32SDK}/i?86-*mingw32* | head -n 1`
>  win_cross=`ls -1 ${WIN32SDK}/bin/*mingw32*-gcc | head -n 1 | sed
> 's/gcc$//'` 
> +test -f ${SYS_PREFIX}/include/WinSock2.h || ln -s
> $win_sysroot/include/winsock2.h ${SYS_PREFIX}/include/WinSock2.h
> 
> Why are you doing this instead of adding another
> `-I$win_sysroot/include` in targets/win32/build-binary#L89?

because I did not find the right spot to do such a thing

at the other hand: I'd rather not need such a shitty hack at all

> Also, does this filename case matter?
 
That's been the issue behind: I cross-compile for Win32 under Linux.
That's when it failed to work.

0-8-15:
Am Wed, 04 Nov 2020 00:45:55 -0800

> 
> 
> 
> > @@ -87,7 +87,7 @@ if [ `is_standalone_app` = "yes" ]; then  
>  else
>    if [ `is_gui_app` = yes ]; then
>      veval "$SYS_CC -I$SYS_PREFIX/include \
> -      $cflag_additions -mwindows win32_microgl.c main.c icon.o -o
> $tgt \
> +      $cflag_additions -mwindows -fpermissive win32_microgl.c main.c
> icon.o -o $tgt \
> 
> Shouldn't the `-fpermissive` also only be applied when using a c++
> compiler, as I thought this was only needed if you use non-conforming
> code?

Likely it should.  Again just a bad work around the actual issue: maybe
someone knows how to fix the offending lines in the source code in a
way compatible with both dialects.

gassechen:
hello. Im try cross compile linux for windows platform.

` checking build system type... x86_64-unknown-linux-gnu`
`checking host system type... i386-pc-mingw32`
`checking target system type... i386-pc-mingw32`
`checking for i386-mingw32-gcc... arg++ -g -O0 -isysroot /usr/mingw-w64-v8.0.0_gcc10.2.0/x86_64-w64-mingw32/ -DWIN32 -DWINVER=0x0501 -fcommon -I/root/.cache/lambdanative/win32/include -L/root/.cache/lambdanative/win32/lib`
`checking whether the C compiler works... no`
`configure: error: in `/root/.lambdanative/tmp_install/gambc-v4_7_9':`
`configure: error: C compiler cannot create executables`
`See `config.log' for more details`

I tried various ways but I can't get it to compile.
how I can help?

mgorges:
@gassechen Sorry is this related to this pull request (which is about linking) or can I move it to a new issue?

gassechen:
yes, can move it to a new issue.

PR #392, opened 2020-11-20
define cond expand feature debug

0-8-15:
This enables compile-time instrumentation/alteration of code depending
on the configured options.

### DIFF
--- a/languages/scm.sh
+++ b/languages/scm.sh
@@ -56,6 +56,7 @@ compile_payload_scm()
     # prep the compiler options
     if [ $SYS_MODE = "debug" ]; then
       scm_opts="(declare (standard-bindings)(extended-bindings)(debug)(debug-location))"
+      scm_opts="${scm_opts}(define-cond-expand-feature debug)"
       payload_cdefs="$payload_cdefs -D___LIBRARY -D___PRIMAL"
     else
       scm_opts="(declare (block)(not safe)(standard-bindings)(extended-bindings))"


mgorges:
Happy to merge this after I tested, but I am wondering if we need to start having more than one debug level as it seems we are creeping towards so much extra stuff that things are getting long to compile and run that somewhat limits the ability to use debug mode in everyday development?

0-8-15:
Am Fri, 20 Nov 2020 10:10:19 -0800

> Happy to merge this after I tested, but I am wondering if we need to
> start having more than one debug level as it seems we are creeping
> towards so much extra stuff that things are getting long to compile
> and run that somewhat limits the ability to use debug mode in
> everyday development?

;-)

+1

My first commit message included a paragraph with the same idea.

Just how this could be done in an extensible and easy to customize way
it currently beyond me.

mgorges:
Is this case sensitive? - The only place I find when I search for this on Google is https://mailman.iro.umontreal.ca/pipermail/gambit-list/2007-May/001368.html which uses `DEBUG` which is also what we use in C?

0-8-15:
Am Sat, 21 Nov 2020 21:09:36 -0800

> Is this case sensitive? - The only place I find when I search for
> this on Google is
> https://mailman.iro.umontreal.ca/pipermail/gambit-list/2007-May/001368.html
> which uses `DEBUG` which is also what we use in C?

Judging from the current code base I would be surprised if we started
suddenly to favor uppercase for Scheme identifier.

With Gambit one can control via the reader settings whether or not
source code is case sensitive.  By default it is.  (Though the original
default for Scheme was not to be.)

Personally I tend to prefer case sensitive symbols and typically write
them in lower case.

Eventually: maybe we can please either take it as is or leave it out
completely.  I'm actually using the lower case version everywhere and I
will not go back and change all the code.

mgorges:
Alright I will refer people to this when the complaints about this being inconsistent start coming in.

PR #394, opened 2020-11-20
libGD: fix dependency issue

0-8-15:


mgorges:
Is this an indirect dependancy? As I only see this being included by apps/DemoAndroidLNjScheme/LIBRARIES but I don't see it being used anywhere else?

0-8-15:
Am Fri, 20 Nov 2020 10:47:37 -0800

> Is this an indirect dependancy? As I only see this being included by
> apps/DemoAndroidLNjScheme/LIBRARIES but I don't see it being used
> anywhere else?

Apparently it is.

I found out while trying to compile, but was side-tracked by events in
between and don't recall all the details.

Summary: Alls I now is: it did not work without and did with.

mgorges:
This adds a system-wide dependancy for all platforms, so needs more testing way beyond me saying, yes it compiles on macosx...

PR #396, opened 2020-11-21
ANDROID: error out when build requires to write system directories

0-8-15:
Along with the fix/warn/advice message it also adds a comment which helps emacs users to maintain the local whitespace policy.

### DIFF
--- a/targets/android/build-binary
+++ b/targets/android/build-binary
@@ -177,8 +177,15 @@ else
       if [ ! -s "$v4file" ]; then
         v4aar=`find $v4dir -name support-v4*.aar`
         if [ -s "$v4aar" ]; then
-          unzip $v4aar classes.jar -d $v4dir > /dev/null
-          v4file=`find $v4dir -name classes.jar`
+          if [ -w $v4dir ]; then
+            echo "WARNING: extracting classes from" $v4aar " in directory $v4dir"
+            unzip $v4aar classes.jar -d $v4dir > /dev/null
+            v4file=`find $v4dir -name classes.jar`
+          else
+            echo Directory $v4dir "is not writable, please run the follwong manually:"
+            echo '"' unzip $v4aar classes.jar -d $v4dir "'"
+            exit 1
+          fi
         fi
       fi
       # If a classes.jar file has been found, copy it to libs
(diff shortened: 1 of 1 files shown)


mgorges:
Could we write these somewhere else instead? Also, is it possible to set this Emacs stuff on a global project-wide basis, so we don't pollute every file with it?

peterlew:
Writing them somewhere else would make sense, but then we just have to check two places when locating it. Any suggestions for a good place to write them?

mgorges:
In case of doubt, I'd write to the lambdanative cache, but I appreciate that needing to check two places might complicate things. Maybe we would need a flag that gets set when we have the need to check in a different place in the config (as we do for the LaTeX test) and then look there instead?

0-8-15:
Am Wed, 02 Dec 2020 11:51:59 -0800

> In case of doubt, I'd write to the lambdanative cache, but I
> appreciate that needing to check two places might complicate things.
> Maybe we would need a flag that gets set when we have the need to
> check in a different place in the config (as we do for the LaTeX
> test) and then look there instead?

The perfect solution is elusive to me.

Checking several places is little fun and complicates code to maintain.

Unconditionally writing to anywhere but cache and local directory is
IMHO to be left to the control of the user.  (That's what the error
would do.)

Smooth workflow does not suddenly ask users to install stuff anywhere.
(The downside of error out.)

However: this toolchain is already in the cache, isn't it?  How about,
if we where to just copy the directory over and then look only in the
cache?

peterlew:
I think a good solution would be to check for the file in the default location, and if it's not found, check a location in the cache. If it's not found there, try to unzip the file to the cache. 

I'm not sure where in the cache would be a sensible place though. Maybe in android/{platform}-{api}/lib ?

peterlew:
After talking with Dustin, we think a good place to put this .jar file would be a new folder called Caches/lambdanative/android/support. There may be multiple .jar files here, one for each API. 

This folder would be removed by `make scrub` but not by `make clean`. 

I will make this change unless there are any objections.

mgorges:
That sounds reasonable to me, but I am also not the one who depends on it?

peterlew:
Resolved by commit 8102b43

PR #398, opened 2020-11-24
ANDROID: allow superclass to see touch events

0-8-15:
As there are complaints about touch/backbutton not doing the expected thing, I recalled that I fixed something related some time ago.  It is required by webview and may help elsewhere.

### DIFF
--- a/loaders/android/bootstrap.java.in
+++ b/loaders/android/bootstrap.java.in
@@ -259,6 +259,7 @@ class  xGLSurfaceView extends GLSurfaceView {
     setRenderer(renderer);
   }
   public boolean onTouchEvent(final MotionEvent event) {
+    super.onTouchEvent(event);
     t=0;
     x=(int)event.getX(); y=(int)event.getY();
     switch (event.getAction()&MotionEvent.ACTION_MASK) {


0-8-15:
I just now recalled why I excluded this little change so far: it's incomplete.

The idea at the moment is:

1. map the native back-key of android to the escape by default
2. have a way to change this mapping
3. pull all the long-pressed events from the widgets code to a generic dispatch
4. make the dispatch configurable
  a. enable several long-press actions (e.g., long-press 1..3 seconds: option $A; 3..6'' option $B; ...)
  b. consider *modifier* keys in this context
  c. enable configuration of the cross product of (a), (b) and the things I'm overlooking by now.

Summary: it's OK for now like this (assuming it does not break things).  But we better add a comment that the particular behaviors is likely to change in the future.

mgorges:
Do you want to add these four things - I accidentally already included the one-line change in 20b33a1

0-8-15:
Eventually I intent to **have** those things.  As I mentioned, it's IMHO OK for now to include it as it is now.  Just expect changes.

PR #399, opened 2020-11-30
ANDROID: Fix setContentView wrt. static declaration by 20b33a1

0-8-15:
The patch fixes the main issue created in the hotfix wrt. `requestAllPersissions` and friends when I discovered that `onCreate` may be called several times per lifetime of an app.

PS: sorry for the removal of offending white space.  I'm busy.  Just delete those random blanks as a drive-by-patch, please.  They do IMHO more harm than good.

### DIFF
--- a/loaders/android/bootstrap.java.in
+++ b/loaders/android/bootstrap.java.in
@@ -1,3 +1,4 @@
+/* LN bootstrap -*- mode: java; c-basic-offset: 2; -*- */
 /*
 LambdaNative - a cross-platform Scheme framework
 Copyright (c) 2009-2020, University of British Columbia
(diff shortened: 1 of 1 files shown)


mgorges:
> Just delete those random blanks as a drive-by-patch, please. They do IMHO more harm than good.

Sorry I am not quite sure what you are asking me here to do - remove them as an edit to your commit before I merge, or accept that this commit removes them and merge this as is? 

I appreciate that the underlying issues are a feature of git, as it prevents us from re-formatting / reflowing code everywhere, but I rely on the ability to use `git blame` to figure out why/how something was changed. Removing even a single space at the end of the line is a change to the respective line, which breaks this behaviour as I then manually need to look at the change, and do a new git blame for the parent, which is not supported on Github and also poor workflow when done on the command line as it breaks my mental model having to integrate code from two places to get a full picture.

mgorges:
Before we embark on something like this https://eev.ee/blog/2016/06/04/converting-a-git-repo-from-tabs-to-spaces/, which uses a git filter in `.gitattributes` if we could force [git blame -w](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt--w) as the default that might also work?

PR #401, opened 2020-12-02
ANDROID: JNI error checking and free local references

0-8-15:
there will be more in audio at least, not yet written.


mgorges:
Sorry, I don't see anything with respect to error checking - only freeing of three references? Did you mean to include something else in this pull request as well?

0-8-15:
Am Wed, 02 Dec 2020 18:33:22 -0800

> Sorry, I don't see anything with respect to error checking - only
> freeing of three references? Did you mean to include something else
> in this pull request as well?

Java error checking in this case, here: `JNI_forward_exception_to_gambit`

mgorges:
But that was there before the commit already?

0-8-15:
Maybe, in which case it is just plugging a memory leak.

https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/jni_refs.html#d8649e76

PR #403, opened 2020-12-02
ANDROID jSCHEME: update following changes to setContentView

0-8-15:


mgorges:
There is some major rearrangement of files in modules/ln_jscheme/ANDROID_java_oncreate which I can't test, but you are the expert for this module, so will take it as is without review.

PR #405, opened 2020-12-02
CORE: remove unused-in-ln, deprecated predicate list-notempty?

0-8-15:
This appears to be not used anywhere.  If it is used elsewhere, it is
better replaced!  Complexity class: O(n)

Suggested replacements:

1. `pair?` - test head only

2. in case of doubt wrt. input: once `list?` to test object being a
propper list and `pair?` once the former succeeded.

### DIFF
--- a/modules/ln_core/list.scm
+++ b/modules/ln_core/list.scm
@@ -37,9 +37,6 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 |#
 ;; list related extras
 
-;; Check for empty lists - this one is better than using pair? for it!
-(define (list-notempty? lst) (and (list? lst) (not (null? lst))))
-
 ;; Make a list of length n, where each element is set to elem
 (define (make-list n #!optional (elem 0))
   (if (zero? n) '()


mgorges:
Well I disagree strongly - just because you might not see it, doesn't mean it won't exist: 35x apps/telePORT/main.scm, 2x apps/VNCaseWriter/main.scm, 10x apps/VitalPAD/main.scm, 1x apps/iControlSrv3/upi.scm; 1x apps/lint/main.scm; 2x plugins/vitalnode/vitalnode.scm are the ones of a very quick grep. So unless we have a replacement that doesn't require me to touch critical code, this will NOT be merged.

0-8-15:
I don't mind either way.

IMHO the predicate SHOULD be deprecated and replaced ASAP anywhere it is being used.  Suggested replacements are given.

The code is just too bad.  If this is part of any critical code, I'd personally tend to withdraw my trust in those apps.

mgorges:
I completely disagree with the trust statement. - Just because you might not like the performance, that doesn't invalidate that this code works and has existed safely in applications that have undergoing hundreds of hours of automated and manual tests. If I now replace them either touching this or the respective apps, it will pull in mountains of unrelated changes, which poses significant safety risks. Finally, why is `(not (null? x))` worse than `(pair x)`?

0-8-15:
Well, the trust statement was, and is, my personal opinion.

I know your safety attitude, which I'd summarize as *"never touch a running system"*.  And I welcome it on those grounds.

Nevertheless I'd still like to strongly suggest to deprecate the predicate and eventually remove it **after having gone to the trouble** to check every single use case and fix it.

The issue is not `(not (null? x))` vs. `(pair? x)`.  While I'd still change in former into the latter for saving one nesting level, the gain (as in runtime) would likely be nothing at all the microseconds.  The other issues are all worse:

1. When `list-notemtpy?` is equivalent to `pair?` then the compiler will inline the `pair?` (at least in LN production mode), the predicate becomes a call to a top level binding, no matter what.
2. Calling an actual procedure is always costlier than inline code, let alone that calling a procedure from another module is like much worse than one defined in the same module, possibly even with local scope.
3. Offering this predicate invites usage, which in turn disables optimizations even in those other modules, where the predicate is used, but since it is not inlineble, results in an inter-module call.  The worst we can have.
4. Asking the question "who would use it", and knowing 1..3 already, I'd say: those who don't.
5. "Why would one use it?", given (4), I'd assume it likely being used by users having read `(= (length lst) 0) ` as a loop-aborting condition.  They might not look at the code while taking computational complexity in mind.  And IMHO `list-notempty?` looks even more innocent than the former, where `length` clearly indicates that the operation is **O(n)** with  being the length of `lst`.
6. When a predicate of class O(n) is used in a loop, the resulting operation is O(n*m) -- already exponential.  This works for a while while there are only very small numbers of elements in each dimension.  (That's where my mistrust origins.)
7. If it was not for (1) than 96a92581 mentions **O(n)** as in (5).  So why?  The `list?` predicate is like `length`, O(n).
8. simply removing the `(list? lst)` from line 41 changes the semantics.

My conclusion: the predicate is broken beyond repair and was better removed.  Even when this means fixing all the app.

NB:  I've been bitten by complexity issues before.  :-/

mgorges:
Most uses are before we go into a loop or call another expensive function; a lot of it are tests on returns of asynchronous communication of data received by [rupi-cmd](https://github.com/part-cw/lambdanative/blob/ceea7d3bfab077f0e1d538eb10c62b73e3a11c32/modules/rupi/rupicore.scm#L141), but I don't think I can share good examples here.

mgorges:
The only public piece I see is in [LNhealth: apps/telePORT/main.scm](https://github.com/part-cw/LNhealth/blob/0a1d0f32862b41d1a5e7506d1b9a1fa6d700f4f2/apps/telePORT/main.scm) but it doesn't illustrate it well.

0-8-15:
> Most uses are before we go into a loop or call another expensive function; a lot of it are tests on returns of asynchronous communication of data received by [rupi-cmd](https://github.com/part-cw/lambdanative/blob/ceea7d3bfab077f0e1d538eb10c62b73e3a11c32/modules/rupi/rupicore.scm#L141), but I don't think I can share good examples here.

Well, this does **not** make the situation better.  However, it is your situation, which looks shaky to me.

I was just about to create a new issue for this.  Will do.  There should be a documented protocol to follow for gradually turning exported bindings from "documented for use" over a few steps into non-existent anymore.  With nagging warnings at first but still working as before, error out in debug mode only, always error out to simply expecting the SEGV once removed and in production mode.

0-8-15:
As for the tests: one of the steps while going though to the trouble of gradually changing the code would be to simply move this line into the tests which use is, including a prominent comment that is is deprecated and than simply leave them as they are.  Just make sure the baseline code used can get rid of it, because it does more harm than good to the code base.

0-8-15:
> I know your safety attitude, which I'd summarize as _"never touch a running system"_. And I welcome it on those grounds.

At the other hand, the `wontfix` label is a bit irritating to me.  I would not bother to submit a PR if this was just for me.  I consider the removal of code which is arguable broken beyond repair an improvement to the code base.  In this case seriously suggesting to go to the trouble to manually assess the consequences and either factor it out literally into those tests using it (here I don't suggest modifying the tests, I'd rather really cut&paste it where required) or otherwise simply use `pair?` instead of `list-notempty?` and in those likely rare cases where you ever actually need to proof that this is a proper list in the first place factor this O(n) code out of the loop.  It improves your code eventually.

And... with the recently extended debug options it is no longer that hard to track issues down as it was before.

Still takes time.  I known.  But `wontfix` is a bit irritating to me.

mgorges:
Fine - I'll change to invalid. As it stands, this pull request is solely to remove a function that is being used, which will cause segfaults. I agree that finding ways to remove it eventually might be desirable, but I won't touch our closed-loop controller for something that might be O(n) or even O(n^2) on a short lists, and where replacements might cause failures in scenarios not covered in current tests.

PR #408, opened 2021-01-16
Improve speed of generalized-arrays module.

gambiteer:
These changes have been tested in both Gambit-C 4.7.9 and Gambit 4.9.2.

These changes are in response to comments by Joerg Wittenberger.

Here we try to speed up the implementation by a number of techniques:

1.  Cache often-accessed properties of intervals (dimension and volume) and arrays (array-elements-in-order?).

2.  Employ fixnum operations by assumeing that the lengths of vectors and the maximum number of arguments to functions are fixnums.  Add an option to the test suite to generate random intervals with bignum limits to test that we didn't use fixnum arithmetic where generic arithmetic is needed.  (This caught a bug in array-tile.)

3.  Newly allocated specialized arrays always store their elements adjacently and in order, so store that knowledge in the array.  If an array has its elements in order, so does a translated version of that array, and so do curried "slices" of that array.

4.  We incorporate a constant table of vectors of zero elements to use in the common case of zero lower bounds of intervals.

generalized-arrays.scm:

1.  *This point does not apply to LambdaNative.*

2.  Introduce fixnum numerical operations by assuming that vector lengths and indices and maximum number of arguments are fixnums.

3.  Make %%every take only a single list (and update the one place in test-arrays that used a multi-list %%every).

4.  Precompute and store the value of %%interval-dimension; cache the value of %%interval-volume after first computation.  Define %%vector-of-zeros, a constant vector of vectors of zeros for common case of zero lower bounds of intervals of dimension <= 4.  Refactor make-interval, add %%finish-interval, which adds %%interval-dimension and %%interval-%%volume fields to intervals.  Change all calls to make-%%interval to calls to %%finish-interval.

5  Change many error checks of, e.g.,

        ((not (exact-integer? i))
         (error "interval-lower-bound: The second argument is not an exact integer: " interval i))
        ((not (< -1 i (%%interval-dimension interval)))
         (error "interval-lower-bound: The second argument is not between 0 (inclusive) and (interval-dimension interval) (exclusive): " interval i))

to

        ((not (and (fixnum? i)
                   (fx< -1 i (%%interval-dimension interval))))
         (error "interval-lower-bound: The second argument is not an exact integer between 0 (inclusive) and (interval-dimension interval) (exclusive): " interval i))

because for the second condition to be true, we need i to be a fixnum, not just an exact-integer.

6.  Change the predicate arguments to %%every and %%vector-every to be closures instead of references to global routines, e.g.,

(define (translation? translation)
  (and (vector? translation)
       (%%vector-every exact-integer? translation)))

to

(define (translation? translation)
  (and (vector? translation)
       (%%vector-every (lambda (x) (exact-integer? x)) translation)))

This is admittedly a Gambit-specific optimization, but changes an intermodule call of a general library routine to an intramodule call of an inlined closure.

7.  Change internal calls of %%interval-lower-bounds->vector to just %%interval-lower-bounds---we don't modify the result, so we don't need a new copy.

8.  Change %%interval-volume to cache the result after computation.

9.  Open-code %%interval=, as it's used a fair amount in error-checking.

10.  Add an in-order? field to a specialized array, which indicates whether the elements of the array are stored adjacently and in order, that can take one of three values: #t, #f, and a non-boolean value %%order-unknown (which we arbitrarily set to 1).

11.  Change calls to (zero? x) to (eqv? x 0) when this is really what we want to check for, because it's more precise.  The is strictly correct, and is inlined to a single instruction by the Gambit compiler.  Same for (= x 1) to (eqv? x 1), etc.

12.  Pull out %%compute-array-element-in-order? from %%array-elements-in-order?.  The new routine takes an interval and an indexer as arguments instead of an array.  %%array-elements-in-order? simply computes and caches the value.

13.  Add in-order? argument to %%finish-specialized-array, pass it to make-%%array.

14.  %%make-specialized-array: new specialized arrays are always in order when first constructed.

15.  Add optional in-order? argument to %%specialized-array-share, which defaults to %%order-unknown, and pass it to %%finish-specialized-array.

16.  Use (%%array-dimension array) instead of (%%interval-dimension (%%array-domain array)).

17.  array-tile: Use generic arithmetic instead of fixnum arithmetic (this is a bug fix).

18.  specialized-array-translate: the "order"ness of the translated array is the same as the original, so pass it to %%specialized-array-share.

19.  Define %%vector-of-trues to use in a one-argument array-reverse for <= 4 dimensions (a small optimization).

20.  %%specialized-array-curry: Precompute whether the subarrays are in-order, wihtout instantiating a subarray, and pass this to %%specialized-array-share when each subarray is created.

21.  array-fold-right: omit second argument to array-reverse.

22.  %%specialized-array-reshape: the reshaped array has elements adjacent and in order if and only if the same is true of the argument.

23.  The code used to compare storage classes with equal?.  Because storage class structures contain functions, if two of them aren't eq? then it's pretty much impossible to compare them, because comparing functions in Scheme is pretty indeterminate.  So I decided to compare storage classes with eq?.

test-arrays.scm:

1.  Adapt to internal changes to generalized-arrays.scm.

### DIFF
--- a/modules/generalized-arrays/generalized-arrays.scm
+++ b/modules/generalized-arrays/generalized-arrays.scm
@@ -136,21 +136,20 @@ OTHER DEALINGS IN THE SOFTWARE.
     (define c128vector-copy! #f))))
 
 
-;;; We need a multi-argument every, but not as fancy as in Olin Shiver's
-;;; list library.  (Shiver's version works fine, though, for our purposes.)
+;; Inlining the following routines causes too much code bloat
+;; after compilation.
 
-(define (%%every pred list . lists)
-  (if (pair? lists)
-      (let loop ((lists (cons list lists)))
-        (or (null? (car lists))
-            (and (apply pred (map car lists))
-                 (loop (map cdr lists)))))
-      (let loop ((list list))
-        (or (null? list)
-            (and (pred (car list))
-                 (loop (cdr list)))))))
+(declare (not inline))
 
-;;; the following is used in error checks.
+;;; We do not need a multi-argument every.
+
+(define (%%every pred list)
+  ;; don't inline %%every, but unroll the loop if advantageous
+  (declare (inline))
+  (let loop ((list list))
+    (or (null? list)
+        (and (pred (car list))
+             (loop (cdr list))))))
 
 (define (%%vector-every pred vec #!optional (vec2 (macro-absent-obj)) #!rest vecs)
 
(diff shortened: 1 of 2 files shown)


mgorges:
test-arrays.scm pass in lambdanative commit 967f8df. The commit message is very long so might just refer here for the details.

PR #418, opened 2021-04-29
EVENTLOOP: ensure exclusive Gambit thread exists.

0-8-15:
At 2020-08-07 Marc Feeley wrote that the situation how lambdanative
uses Gambit (at least on Android) is not supported by Gambit, he
wrote:

> Any number of Scheme threads can call C functions, but only one
> Scheme thread at a time can do a Scheme to C call that calls back to
> Scheme.  This is because the C stack is shared by all the Scheme to
> C calls... if the Scheme threads T1 and T2 call C (in the order T1
> then T2) and if T1 returns to Scheme while T2's call is still in
> progress, then T2's C stack frame will be removed when T1's C stack
> frame is removed (because T2's frame was added after T1's frame).

Under Android both the UI thread and the OpenGL thread call into
Gambit; all bets are off.  Race conditions are not avoidable.  This
seems to explain random crashes I observed eventually to the point
that I could reproduce them by tapping the screen fast enough.

The only reasonable escape I could conceive was to use Gambit as Marc
intented: one single thread running contineously.

The upshot is: for Android and X11 based systems running Gambit from a
dedicated thread avoids jitter and event loop related delays for
Gambit threads.  In practice this work much smoother.

## Status

This change will certainly break some applications:

1. At least those, where race conditions are exacerbated to sure
   conflicts.
2. A batch of untested modifications required for compatible with this
   patch is half prepared already.  Notably the 'serial' module is
   unlikely to work out of the box.
3. I assume that those other X11 based system should mostly work.  Non
   tested, non available here.

This change avoids many C-to-Scheme calls, which where not properly
protected and runs the event loop in a pthread of its own for X11 and
Android.  Under win32 it still calls the previos way, but used
EVENT_IDLE to better wait with the Gambit thread system than block
threads in Windows.

**For Android the change includes a new NativeGLSurfaceView and
supporting GLState class.  These work so far more reliable than the
version before.  But the are premature.  Frequently the android app
will come up first time after installation but not again.  Forced stop
of the app and restart helps.  Further restarts magically work.**

Tested are so far on Android, Linux, Windows.

### DIFF
--- a/LNCONFIG.h.in
+++ b/LNCONFIG.h.in
@@ -117,10 +117,19 @@ int  scm_screenheight();
 
 // hook prototype (performs scheme init)
 void ffi_event(int,int,int);
+void ln_gambit_unlock();
+typedef struct {
+  int t;
+  int x;
+  int y;
+  int fd[2];
+} ffi_event_params_t;
+extern ffi_event_params_t ffi_event_params;
 
 // C prototypes
 // these are used by the desktop platforms
 void microgl_hook(int,int,int);
+void microgl_draw_before();
 void microgl_swapbuffers();
 void microgl_close();
 void microgl_init();
@@ -130,6 +139,7 @@ void microgl_pollevents();
 void microgl_refresh();
 int microgl_screenwidth();
 int microgl_screenheight();
+unsigned int microgl_redraw_period(unsigned int);
 
 #endif // STANDALONE
 
--- a/languages/scm.c
+++ b/languages/scm.c
@@ -42,6 +42,9 @@ void lambdanative_payload_setup()
   setup_params.debug_settings = debug_settings;
   lambdanative_exit_call_count = 0;
   ___setup(&setup_params);
+ /*  Runing from a single thread should fix the race conditions, which
+     lead to the following mitigation.  Should be ready to be removed.
+
   #if defined(ANDROID)
     #if (___VERSION < 409002)
       ___disable_heartbeat_interrupts();
@@ -49,6 +52,7 @@ void lambdanative_payload_setup()
       ___cleanup_heartbeat_interrupt_handling();
     #endif
   #endif
+ */
 }
 
 void lambdanative_payload_cleanup()
--- a/libraries/libgambit/0001-Fix-incorrect-space-allocation-report-by-time-specia.patch
+++ b/libraries/libgambit/0001-Fix-incorrect-space-allocation-report-by-time-specia.patch
@@ -1,0 +1,50 @@
+From e529b1d6d5b7bae2fcc3e9cd8e7f3e11a1318193 Mon Sep 17 00:00:00 2001
+From: Marc Feeley <feeley@iro.umontreal.ca>
+Date: Sun, 28 Mar 2021 19:29:54 -0400
+Subject: [PATCH] Fix incorrect space allocation report by time special form
+
+---
+ lib/_kernel.scm | 6 ++++--
+ lib/mem.c       | 3 ++-
+ 2 files changed, 6 insertions(+), 3 deletions(-)
+
+diff --git a/lib/_kernel.scm b/lib/_kernel.scm
+index 59b9e41..2d827dd 100644
+--- lib/_kernel.scm
++++ lib/_kernel.scm
+@@ -4346,7 +4346,9 @@ end-of-code
+
+     if (!___FIXNUMP(result))
+     {
+-      n = ___bytes_allocated (___PSPNC) - n;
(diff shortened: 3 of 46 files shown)


mgorges:
'EVENTLOOP: ensure exclusive Gambit thread exists.' is where the actual changes are, correct - everything else are duplicates of things in #417 ? Breaking serial will be a big problem for us as we rely on this. Also, I am personally concerned about breaking clinical data collection systems which use Linux (without GUI) and which will cause irrecoverable data loss if things go bad.

mgorges:
Can't say that I even understand 20% of this new code but I have a few questions/comments:

1. You disable `glgui-timings-at-10msec!`, which we use in a lot of apps, including all apps (including non-gui) that use the [modules/scheduler/scheduler.scm](https://github.com/part-cw/lambdanative/blob/74ec19dddf2f2ff787ee70ad677bc13b9dfafc29/modules/scheduler/scheduler.scm) module and need proper timing? The minimum as per https://github.com/part-cw/lambdanative/pull/418/files#diff-b3b574bd3e4985858199598c697225c69645bbe5420cc7a4402f8e62f5625b04R268 seems to be 0.05 sec sleeps now?
2. As for your question where `glCoreTextureReset` is used - it is in `glCoreInit`
3. You disable the `EVENT_IDLE` in `event_timer_callback` on win32, why? - I think win32 has this annoying behaviour if a GUI app is active but not in focus the main loop will stop running after ~30 seconds (or maybe a minute, can't recall), and when you return to it it throws lots of events and tries to catch up which messes up the output plugins in the scheduler, but this mitigated as introduced in fdc3626e ?
4. The wrong event code (127 instead of 128) in `ANDROID_JAVA_ONDESTROY` in https://github.com/part-cw/lambdanative/pull/418/files#diff-2098455a1eec986f28edb431eaec9868112f3f3dc606d46435c07066363a5e11R287 might explain some strange app closing behaviour on Android. Nice catch!
5. We still need a way to ensure all the other Android things like camera, serial, audio, etc work and I am not sure how to do that short of building this branch of LN and waiting for things to break?

0-8-15:
Am Wed, 26 May 2021 20:43:33 -0700
schrieb Matthias Görges ***@***.***>:

> Can't say that I even understand 20% of this new code but I have a
> few questions/comments:
> 
> 1. You disable `glgui-timings-at-10msec!`, which we use in a lot of
> apps, including all apps (including non-gui) that use the
> [modules/scheduler/scheduler.scm](https://github.com/part-cw/lambdanative/blob/74ec19dddf2f2ff787ee70ad677bc13b9dfafc29/modules/scheduler/scheduler.scm)
> module and need proper timing? The minimum as per
> https://github.com/part-cw/lambdanative/pull/418/files#diff-b3b574bd3e4985858199598c697225c69645bbe5420cc7a4402f8e62f5625b04R268
> seems to be 0.05 sec sleeps now?

OK, so we need to bring that back.

There is no reason to have a 0.05 minimum.

At the other hand: I don't understand what this scheduler is good for
and supposed to do.

Personally I have a hard time to envision any need for a 100Hz poll
loop. Even for visual feedback it is too fast, for audio it is too slow
(let alone that there is usually specialized sound hardware these days).

Perhaps there is a better solution to the problem solved by that
scheduler, but as I said: we probably better simply bring it back.

> 2. As for your question where `glCoreTextureReset` is used - it is
> in `glCoreInit`

Thanks.  So if that's the only case, great!

> 3. You disable the `EVENT_IDLE` in  `event_timer_callback` on win32,
> why? - I think win32 has this annoying behaviour if a GUI app is
> active but not in focus the main loop will stop running after ~30
> seconds (or maybe a minute, can't recall), and when you return to it
> it throws lots of events and tries to catch up which messes up the
> output plugins in the scheduler, but this mitigated as introduced in
> fdc3626e ?

Because it looked to me as a useless piece of code which I did not
understand.  So I wanted to see what breaks when I disable it.  Now I
copy&pasted you comment above into `event_timer_callback` in order to
avoid loss of that knowledge.

> 5. We still need a way to ensure all the other Android things
> like camera, serial, audio, etc work and I am not sure how to do that
> short of building this branch of LN and waiting for things to break?

Unfortunately I don't have a better suggestion to find out.

For serial I'm pretty sure it will break.

Let me repeat my offer: I'm ready to help fixing the breakage as soon
as I can either understand or repeat it.

For serial, this means I need to understand how I could set up a test
case or learn what is is supposed to do.  (That is: there might be an
alternative, which could perhaps be even better: Android is old Linux
under the hood.  Maybe it's possible to use native device access?  I
dunno.)

PR #424, opened 2021-09-24
[FIX][DOWNLOAD] fixes logic for checking header 

karliwalti:
When header was 'false' the logic comparisons failed. 

Also: when file can't be downloaded drop a warning instead of a status log


mgorges:
Thanks for the proposed fix - we might need to ask @ddunsmuir to check this won't break anything for him, but he is currently focusing on clinical deployments in the field.

ddunsmuir:
I don't have any active apps using this module and also the change makes sense to me and looks good.

PR #435, opened 2022-10-03
M1/M2 (apple silicon) macosx support

paines:
Little patch to support m1/m2 (apple silicon) Macs.

I testest with DemoSchemeAndC (needs brew install autoconf) and Calculator (need brew install imagemagick)

### DIFF
--- a/configure
+++ b/configure
@@ -127,7 +127,7 @@ case "$hostsys" in
 *-linux-gnu*)
   SYS_HOSTPLATFORM="linux"
 ;;
-x86_64-apple-darwin*)
+*-apple-darwin*)
   SYS_HOSTPLATFORM="macosx"
 ;;
 *-openbsd*)


mgorges:
@ddunsmuir can you test this as I don't have an Apple silicone Mac. The same change might also be needed in libraries/libgambit/make.sh?

ddunsmuir:
@mgorges I don't have access to at Apple silicone Mac either, having purchased the last model of Intel ones.

mgorges:
@paines could you check what change might be needed to _libraries/libgambit/make.sh_ I presume we need an extra **if** or **switch** to test for `aarch64-apple-darwin` or `x86_64-apple-darwin` and modify [libraries/libgambit/make.sh#L14](https://github.com/part-cw/lambdanative/blob/master/libraries/libgambit/make.sh#L14). Note that this newer version of GambitC is only active if you modify `modules/config/LIBRARIES` to include `libgambit` instead of `libgambc`

paines:
Actually I am confused by this. I tried both options you mentioned, and then arm64 , without EXTRACONF at all or EXTRACONF=empty and they all failed. BUT If I pull https://github.com/gambit/gambit/archive/v4.9.2.tar.gz and configure --enable-single-host and make it, it simply compiles (independent of the 4 patches in the libraries/libgambit directory).

mgorges:
The line I am pointing to is targeting iOS, not building for macOS, which I believe you did manually. The macOS use case should be in [libraries/libgambit/make.sh#L33](https://github.com/part-cw/lambdanative/blob/master/libraries/libgambit/make.sh#L33), which leaves `EXTRACONF` empty. If you could provide more details of the problem I'd be happy to explore more, but in the absence of an Apple Silicon mac, we currently can't test this.

paines:
@mgorges Sorry about the delay... I uploaded the error here: https://gist.github.com/paines/b69881894304745d722f421504b9e3a3

Like mentioned previously I tried line 34 EXTRACONF parameter of libgambit/make.sh empty or with --host=arm64 --host=aarch64-apple-darwin all with the same error.

mgorges:
That is an unrelated problem - you are changing the compiler version, which makes an inconsistent state with the libraries and header files, i.e. the `/Users/aneslihovac/Library/Caches/lambdanative/macosx/include/gambit-not407009.h:2:2: error: "You are probably compiling C code generated by a Gambit compiler earlier or later than v4.7.9"
#error "You are probably compiling C code generated by a Gambit compiler earlier or later than v4.7.9"` thus if you want to switch from libgambc to libgambit you need to flush the entire lambdanative cache.

paines:
After cleaning the cache, DemoHelloWorld builds and runs fine with libgambit and EXTRACONF=<empty>.

PR #442, opened 2023-10-10
Bluetooth Module

peterlew:
Added a Bluetooth module. Can currently start and stop a scan, and get the info about the results as a JSON vector. Android only - other platforms return empty vector.


mgorges:
@peterlew did you ever look at https://github.com/mgorges/lambdanative-testing/tree/master/modules/bluetooth ?

peterlew:
@mgorges I didn't know you'd done work on this! This module is currently more for scanning than passing information through connections, but it could be very useful to add some of the functionality of your module in the future

mgorges:
Could you look into reapplying or checking if some of those features can simply be added in, please?

Issue #446, opened 2024-01-23
Use apksigner corresponding to Android version specified in configuration


bkmgit:
The first matching SDK is used in:
https://github.com/part-cw/lambdanative/blob/master/targets/android/build-binary#L430
which may not correspond to the Android version used for building when multiple SDKs are installed.
This may lead to an older/newer apksigner program being used and lead to an incompatibility.

mgorges:
I am sorry, but this issue and pull request don't quite make sense to me - the `$ANDROIDSDK` is user-determined in the SETUP, let's say `ANDROIDAPI=24` or `ANDROIDAPI=21`. However, the version of the build-tools is fixed from its installation by the sdkmanager, e.g. `build-tools;30.0.2`. 
As such adding `ls -1 $ANDROIDSDK/build-tools/$ANDROIDAPI*/zipalign` or `$ANDROIDSDK/build-tools/$ANDROIDAPI*/apksigner` will result in neither tool to be found anymore? To illustrate:
```
[...]
 => zipaligning..

/usr/local/android-sdk-macosx
24
/usr/local/android-sdk-macosx/build-tools/30.0.2/zipalign
ls: /usr/local/android-sdk-macosx/build-tools/24*/zipalign: No such file or directory
[...]
```
If you truly must set a specific version of the `build-tools` to use, we must introduce a new variable in `SETUP` for it, but I assume most users only have one set of tools. The other alternative is to hardcode it?

bkmgit:
Thanks. It would probably be better to have an optional variable in `SETUP` or users who may have more than one set of tools.

mgorges:
I saw the updated pull request, but could you comment a bit more on why this is needed; we have always used the latest `build-tools`, and it works for any `$ANDROIDAPI` target as they are backwards compatible. Does that break something specifically for you? Thanks!

bkmgit:
Without this it breaks for me on Linux.  Typically need to build for a variety of APIs, have build tools from 26.0.1 onwards. With the wild card, it sorts numerically in increasing order, so picks up the oldest version of `build tools`.  The software is still functional, but I would need to modify the signing portion of the script.  I may have an unusual setup, not sure how many others have several versions of `build tools` which would justify the extra portions in the scripts.

mgorges:
Thanks - could you try if you only kept the latest version of the `build-tools` if that worked for all APIs you build for? We target different APIs frequently and, as I mentioned, only use a single `build-tools` version universally.

[...]

PR #447, opened 2024-01-23
Choose build-tools corresponding to selected ANDROIDAPI

bkmgit:
Closes https://github.com/part-cw/lambdanative/issues/446


mgorges:
Please see comments in https://github.com/part-cw/lambdanative/issues/446 of why I don't think this makes sense to add.

PR #477, opened 2026-05-09
Add support for libwebp

bkmgit:
Add support for libwebp in libgd and also add a webp module similar to the png module.


mgorges:
Sorry, but it seems that your newly added [modules/webp/webp.scm](https://github.com/part-cw/lambdanative/pull/477/files#diff-152d571d39e722b7606b56bb4494c40377eb58d68a93faf82013240fc73eac6b) is still a PNG version, both in terms of function names and content, so that needs to be revised first before we can merge it.

bkmgit:
Updated the module and added a demo example.  In the demo example, the image is reversed for some reason, but otherwise seems ok.

Can follow up with support for animations, but will experiment with libgd to see if can produce an example for this using gifs first and see if need to additionally add support for [giflib](https://giflib.sourceforge.net/).

mgorges:
This one is a bit hard to test as I can't build GD for iOS (it's fairly old, and the macOS architecture detection check fails `checking host system type... Invalid configuration `arm64-apple-darwin25.5.0': machine `arm64-apple' not recognized`.

bkmgit:
Ok. My understanding is that GitHub actions can be used to build iOs apps:
- https://www.andrewhoog.com/posts/how-to-build-an-ios-app-with-github-actions-2023/
- https://nowham.dev/posts/github-actions-setup-ios/
- https://github.com/tpoechtrager/cctools-port

[xtool](https://xtool.sh/) could also be helpful.

I do not have an Apple developer account though. Tools such as:
- https://github.com/Dadoum/Sideloader
- https://altstore.io/
provide alternative ways to distribute and install iOs applications.

mgorges:
Sorry but the issue is that GD 2.3.3 doesn't compile with my current OS and toolchain, and there is no newer release, not the lack of integrated continuous integration, which you propose to add in #467, for it.

bkmgit:
The continuous integration so far is just Linux and Android builds.  I have not tried iOs builds.  What version of iOs would you want to test it on?  Need to see if the toolchains above can work with objective-c in addition to swift.

PR #478, opened 2026-05-18
WIP: Add basic support for Tiff images

bkmgit:
Tiff is helpful since it allows one to add geo tags using [GeoTIFF](https://en.wikipedia.org/wiki/GeoTIFF) so can enable integration with apps that take pictures or use maps and also incorporate GPS information. Camera app works, but some additional checks maybe needed since the TIFF file format can have very large images and uses uint32 types that maybe bigger than some of the corresponding scheme types.


bkmgit:
This builds on #477 though could be reviewed independent of it. Code changes are similar.

Writing tiff file follows the tutorial at:
https://pages.graphics.cs.wisc.edu/559-f99/libtiff_tutorial.htm

PR #481, opened 2026-05-27
Add demonstration program to scan a QR code.

bkmgit:
To get the program to work, had to make some changes to the [quirc module](https://github.com/part-cw/lambdanative/tree/master/modules/quirc).  Not clear if anything else uses this module.

Some modifications maybe needed for Kanji and other non-ascii QR codes.


bkmgit:
Maybe also useful to adapt it to use [libzbar](https://github.com/mchehab/zbar) or [libzxingcpp](https://github.com/zxing-cpp/zxing-cpp) which cover a wider variety of QR and bar codes.

mgorges:
This is used in one private app (PhoneOx), and I'd have to check whether the removal of `(if (file-exists? pngfile)` may cause any issues. Otherwise will try to review this over the weekend.

bkmgit:
Returned check for png file.

apps/DemoQrscan/main.scm
```diff
@@ -0,0 +1,125 @@
...
+    (make-window 320 480)
+    (glgui-orientation-set! GUI_PORTRAIT)
+    (set! gui (make-glgui))
+    (let ((w (glgui-width-get))
+          (h (glgui-height-get)))
+      (set! background (glgui-pixmap gui 0 0 default:background w h))
+      (let* ((bw 150) (bh 50)
+            (bx (/ (- w bw) 2.))
+            (by (/ (- h bh) 2.)))
+	(set! qr-display (glgui-label gui bx (- by (* bh 2)) bw bh "" ascii_18.fnt Black))
```
ddunsmuir: I find the text is cut-off and doesn't fit your label. I recommend you make qr-display much wider than the buttons and possibly use glgui-label-wrapped instead to allow multiple lines. QR codes are often used for URLs or content that is long to save time by scanning instead of typing.

modules/quirc/quirc.scm
```diff
@@ -137,13 +139,27 @@ end-of-c-declare
...
+               ret
+               (begin
+                 (if (= QUIRC_SUCCESS (quirc_extract_and_decode q (length ret)))
+                   (append ret (list (quirc_decoded_data)))
+                   (append ret (list "Decoding failed")))))))
+          )
+      (quirc_destroy q)
+      result
+      )
+    (list "Image file is missing")
```
mgorges, line 162: This could return #f, which it did before?

apps/DemoQrscan/main.scm
bkmgit: Thanks for the feedback.

modules/quirc/quirc.scm
bkmgit: Thanks will update.

mgorges: The PR is already merged?

PR #484, opened 2026-05-29
BUILD: Enable building packages using CMake

bkmgit:
Add option to  build libraries using CMake.


bkmgit:
Partially addresses #480

PR #486, opened 2026-06-18
WIP: Tinymaix

bkmgit:
Example showing how to use CMake to build a library.

The [TinyMaix library](https://github.com/sipeed/TinyMaix) can probably be integrated, though the bindings may need to not use code derived from TinyMaix to be incorporated.

The current example uses a network trained on the [MNIST](https://en.wikipedia.org/wiki/MNIST_database) dataset - licensing of this is unclear, though it is an open dataset.  This likely cannot be integrated.  Perhaps train a network on the [Dry Bean dataset](https://doi.org/10.24432/C50S4B)?


bkmgit:
Builds and runs, but recognition rate is poor, see https://github.com/sipeed/TinyMaix/issues/93 .  May need to use alternate model.

bkmgit:
@mgorges  @ddunsmuir Would appreciate an initial review.  There is an extended MNIST dataset which one could use to train a more accurate model.  The present model works when one takes a picture of a single number on a white background.  The number should be centered in the image.  One could use a [Maximally Stable Extremal Regions]() to extract the location of the digit, either programming it directly or using a library such as [OpenCVMobile](https://github.com/nihui/opencv-mobile) or [VLfeat](https://www.vlfeat.org), at the moment cropping is used to extract a square region in the middle of the image.

